PR #65, opened 2021-04-08
Api 6054/deploy staging tests

ediiotero:
[API-6054](https://vajira.max.gov/browse/API-6054)

Update package json and smoke-test.sh, regression-test.sh


alexandervalencia:
@ediiotero does this still need to be reviewed/updated or can it be closed?

PR #90, opened 2021-06-17
Only transform securities if api uses apikey security

RockslideJr:
[API-7867](https://vajira.max.gov/browse/API-7867)

This updates the oas transformer to check if an api uses apikey securities before doing security transformations. If it does not, the transformations are skipped entirely.


test/oas/oas.service.test.ts
```diff
@@ -439,7 +464,7 @@ describe('OASService', () => {
         const serverError: ServerError = (err as unknown) as ServerError;
 
         expect(serverError.action).toEqual(
-          'Updating model { name: gimli iteration: 1 }',
+          'Updating model { name: gimli iteration: 0 }',
```
michaelkimball, line 467: What caused this change?

RockslideJr: I'm not entirely sure why it was ever 1 to begin with. My guess is there was some state leakage between tests. Unless we specifically call the save method somehow (which we don't in this test), the iteration should be at 0

michaelkimball (approved):
LGTM 🚀

PR #92, opened 2021-07-06
Split API into separate models

RockslideJr:
This is the first of 2 PRs towards adding the metadata model. This one splits the API model into 2 separate models, API and APIVersion. This split requires touching just about everything, so there's a lot to it.


src/api/version/apiVersion.model.ts
```diff
@@ -0,0 +1,103 @@
...
+    orphanRemoval: true,
+  })
+  oasCounter: CounterModel;
+
+  @OneToMany({
+    entity: () => OASModel,
+    mappedBy: (spec) => spec.apiVersion,
+    orphanRemoval: true,
+  })
+  openAPISpecs = new Collection<OASModel>(this);
```
christineyost, line 65: Is the owner option not needed here?

src/oas/oas.filter.ts
```diff
@@ -0,0 +1,10 @@
+import { OASResponseBody } from './oas.interface';
+import OASModel from './oas.model';
+
+const filterOASResponse = (model: OASModel): OASResponseBody => ({
+  spec: model.spec,
+  transformedSpec: model.transformedSpec,
+  iteration: model.iteration,
```
christineyost, line 7: I think we should include status here too

src/oas/oas.interface.ts
```diff
@@ -270,4 +270,10 @@ export interface OASUpdateParams {
   status?: StatusEnum;
 }
 
+export interface OASResponseBody {
+  spec: OAS;
+  transformedSpec: OAS;
+  iteration: number;
```
christineyost, line 276: I think this should also include status

christineyost (commented):
Nice work on a super big change! I've reviewed all the changes in `src` but haven't taken a look at test changes yet

alexandervalencia (commented):
Looks good so far, I left a few comment to address. Thanks @RockslideJr!

test/api/gatewayConfig/gatewayConfig.filter.ts
```diff
@@ -0,0 +1,16 @@
+import filterGatewayConfigResponse from '../../../src/api/gatewayConfig/gatewayConfig.filter';
+import GatewayConfigModel from '../../../src/api/gatewayConfig/gatewayConfig.model';
+
+describe('filterGatewayConfigResponse', () => {
+  it('strips out the id', () => {
```
christineyost, line 5: This test may suffer the same issue that @alexandervalencia mentioned for the `api.filter` test

test/oas/oas.model.test.ts
```diff
@@ -13,15 +15,16 @@ describe('OASModel', () => {
       security: { apiKey: [] },
     } as unknown) as OAS;
 
-    const api = new APIModel(
-      'the-fellowship',
-      'rivendell',
+    const api = new APIModel('the-fellowship');
+    const apVersion = new APIVersionModel(
```
christineyost, line 19:
```suggestion
    const apiVersion = new APIVersionModel(
```

test/oas/oas.model.test.ts
```diff
@@ -13,15 +15,16 @@ describe('OASModel', () => {
...
+      { serviceId: 'rivendell' },
       'http://Nine-Walkers-shall-be-set-against-the-Nine-Riders-that-are-evil.',
       1,
       SecurityTypeEnum.API_KEY,
+      api,
     );
 
-    const original = new OASModel(oas, transformedOAS, api, 'one_ring');
+    const original = new OASModel(oas, transformedOAS, apVersion, 'one_ring');
```
christineyost, line 27:
```suggestion
    const original = new OASModel(oas, transformedOAS, apiVersion, 'one_ring');
```

test/oas/oas.controller.test.ts
```diff
@@ -15,8 +17,12 @@ describe('OASController', () => {
       majorVersion: 0,
     },
     api: {
+      name: 'the-black-gate',
+    },
+    apiVersion: {
       gatewayId: 'the-black-gate',
```
christineyost, line 23:
```suggestion
      gatewayConfig: {
            serviceId: 'the-black-gate',
            routeId: 'the-black-gate',
      },
      securityType: SecurityTypeEnum.API_KEY,
```

christineyost: Doesn't matter for this test, but might be nice to keep this consistent with our data model

test/api/api.filter.test.ts
```diff
@@ -0,0 +1,12 @@
...
+import filterAPIResponse from '../../src/api/api.filter';
+
+describe('filterAPIResponse', () => {
+  it('strips out the id', () => {
+    const apiModel = new APIModel("shelob's lair");
+
+    expect(filterAPIResponse(apiModel)).toEqual({
+      name: "shelob's lair",
+    });
+  });
```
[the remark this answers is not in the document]
christineyost: Or maybe just manually set the id like we do in other tests

christineyost (commented):
Finished reviewing tests as well

src/api/api.service.ts
```diff
@@ -1,94 +1,52 @@
...
-    name: string,
-    majorVersion: number,
-  ): Promise<APIModel | null> {
+  public async getAPI(name: string): Promise<APIModel | null> {
     return await this.entityManager.findOne(
       APIModel,
-      { name, majorVersion },
-      { populate: ['counter', 'gatewayConfig'] },
+      { name },
+      { populate: ['versions.oasCounter', 'versions.gatewayConfig'] },
```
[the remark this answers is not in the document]
michaelkimball: I feel like this should return at least the relevant versions inside the object if not the OAS

src/api/version/apiVersion.filter.ts
```diff
@@ -0,0 +1,16 @@
+import filterGatewayConfigResponse from '../gatewayConfig/gatewayConfig.filter';
+import { APIVersionResponseBody } from './apiVersion.interface';
+import APIVersionModel from './apiVersion.model';
+
+const filterAPIVersionResponse = (
+  model: APIVersionModel,
+): APIVersionResponseBody => {
+  return {
+    gatewayConfig: filterGatewayConfigResponse(model.gatewayConfig),
```
michaelkimball, line 9: Maybe it's just me, filter implies the removal of items from an array. This is more of a transformation, but take it or leave it.

migrations/Migration20210706170942.ts
```diff
@@ -0,0 +1,12 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20210706170942 extends Migration {
```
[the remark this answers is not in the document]
RockslideJr: I'm not sure why it generated the 
```js
    this.addSql('alter table "oas" drop constraint if exists "oas_api_version_id_check";');
    this.addSql('alter table "oas" alter column "api_version_id" type int4 using ("api_version_id"::int4);');
 ```
lines for each (especially because i'm pretty sure it's not doing anything), but each migration also has a third line that's different

File: src/api/api.service.ts (hunk 9 above)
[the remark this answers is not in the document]
Comment (RockslideJr) [reply src/api/api.service.ts:40]: I'd rather populate here (where it can do it all in one query) for future uses. I don't think we are ever pulling enough data for the extra joins to matter performance wise (which also doesn't really matter since it's just us using it)

File: src/api/api.service.ts [hunk 12]
```diff
@@ -1,94 +1,52 @@
...
     try {
       return await this.entityManager.findOneOrFail(
         APIModel,
-        {
-          name,
-          majorVersion,
-        },
-        { populate: ['counter', 'gatewayConfig'] },
+        { name },
+        { populate: ['versions.oasCounter', 'versions.gatewayConfig'] },
```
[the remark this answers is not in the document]
RockslideJr: Same thought as above

src/api/gatewayConfig/gatewayConfig.interface.ts
```diff
@@ -1,9 +1,9 @@
 export interface GatewayConfig {
   serviceId: string;
-  routeId: string;
+  routeId?: string;
```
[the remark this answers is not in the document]
RockslideJr: ah that makes sense. I thought current state was the long term plan. I'll remove this.

src/api/version/apiVersion.model.ts
RockslideJr: not for `OneToMany`. The owner is always the Many side

src/api/version/apiVersion.interface.ts
```diff
@@ -0,0 +1,21 @@
...
+  majorVersion: number;
+  securityType: SecurityTypeEnum;
+}
+
+export interface APIVersionResponseBody {
+  gatewayConfig: GatewayConfigResponse;
+  oasUrl: string;
+  majorVersion: number;
+  securityType: SecurityTypeEnum;
+}
```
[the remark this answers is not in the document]
RockslideJr: True. I think it's a little clearer to just keep the duplication for now. That way we don't need to keep a long list of strings for things we're omiting

test/api/api.filter.test.ts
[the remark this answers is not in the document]
RockslideJr: good catch. I'm going to instead manually set the id. These really should be unit tests

src/api/version/apiVersion.filter.ts
RockslideJr: my thought was that since the purpose is just to filter out the properties we _don't_ want users to see, the word filter made sense. I'm not attached to it though if you have a better idea

michaelkimball: That happens to be the end result, but not necessarily the purpose. You are transforming a database entity into a response body. What shape or modifications you end up making to the database entity may change over time.

test/api/version/apiVersion.controller.integration.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    describe('DELETE /docserver/apis/:name/versions/:majorVersion', () => {
+      it('deletes the APIVersion and returns a 204', async () => {
+        const response = await request
+          .delete('/docserver/apis/one_ring/versions/0')
+          .auth(process.env['ADMIN_USERNAME'], process.env['ADMIN_PASSWORD']);
+
+        expect(response.status).toEqual(204);
+
+        const createdVersion = await orm.em.findOne(APIVersionModel, {
```
christineyost, line 221:
```suggestion
        const deletedVersion = await orm.em.findOne(APIVersionModel, {
```

test/api/version/apiVersion.controller.integration.ts
```diff
@@ -0,0 +1,246 @@
...
+    describe('PUT /docserver/apis/:name/versions/:majorVersion', () => {
+      it('updates the APIVersion and returns a 200', async () => {
+        const response = await request
+          .put('/docserver/apis/one_ring/versions/0')
+          .auth(process.env['ADMIN_USERNAME'], process.env['ADMIN_PASSWORD'])
+          .send({ ...apiVersion, oasUrl: 'http://www.oas.website' });
+
+        expect(response.status).toEqual(200);
+
+        const createdVersion = await orm.em.findOne(APIVersionModel, {
```
christineyost, line 187:
```suggestion
        const updatedVersion = await orm.em.findOne(APIVersionModel, {
```

christineyost (commented):
Looks really good to me! The only other suggestion I have is to update the readme so that the sample curl commands are up-to-date - or if not now create a ticket to update it.

src/api/version/apiVersion.filter.ts
RockslideJr: I looked at it as the purpose. We have our representation of the entities our API deals with, and we need to filter out the things we don't want people to see before we return it as a response. But I will not be maintaining this, so hit me with a better name

michaelkimball (approved):

alexandervalencia (dismissed):
I feel good about where this is at, thanks @RockslideJr  🚀

alexandervalencia (approved):
♻️

src/api/version/apiVersion.filter.ts
dsteeve: refineGatewayConfigResponse (my $.02)

PR #94, opened 2021-07-06
add a dockerignore for node_modules and dist

wyattwalter:
This makes build times much faster locally if you do an `npm install` outside of the Docker containers as it doesn't have to send ~600M of context to the Docker daemon.


wyattwalter:
Without this change on a container rebuild in Codespaces, I get:

```
real    0m18.961s
user    0m1.415s
sys     0m1.677s
```

With it, I get:

```
real    0m1.050s
user    0m0.048s
sys     0m0.030s
```

christineyost (approved):
Looks good to me! ✨

michaelkimball (approved):

PR #96, opened 2021-07-15
API-8069 Set gatewayConfig routeId to not null

michaelkimball:
[API-8069](https://vajira.max.gov/browse/API-8069)﻿


ediiotero (approved):
LGTM

PR #98, opened 2021-08-04
Api 8183/metadata model

christineyost:
This PR is for [API-8183](https://vajira.max.gov/browse/API-8183). It adds the new MetadataModel which will be used to create the metadata json blob required by Developer Portal for each API. It also updates the existing API and APIVersion models to support this new model.

A few additional notes:
- I left the healthcheck property out for now since the Developer Portal does not make use of this property. If we decide it's needed later we can add it at that point.
- I defaulted `internal_only` to `true` and `status` to `Current Version` for each APIVersion. We will need to add tickets to updated these properties as appropriate for each APIVersion.
- I plan to update the README sample curl commands once changes are agreed upon.


jest.config.js
```diff
@@ -8,11 +8,14 @@ module.exports = {
     },
   },
   coverageThreshold: {
-    '**/*.ts': {
+    global: {
       statements: 80,
     },
+    './src/api/api.model.ts': {
+      statements: 75,
+    },
```
christineyost, line 16: api.model.ts has the following lines uncovered: 20-21,27,34-35

I'm not sure how we can cover these lines. Since this file is so small, this drops the percent of lines covered to 79.17, so I lowered the threshold here to 75%. Open to other solutions to this problem though!

test/oas/oas.service.integration.ts
```diff
@@ -20,6 +20,9 @@ import { DifferService } from '../../src/differ/differ.service';
 import APIVersionService from '../../src/api/version/apiVersion.service';
 import APIVersionModel from '../../src/api/version/apiVersion.model';
 import SecurityTypeEnum from '../../src/api/security/securityType.enum';
+import StatusEnum from '../../src/status/status.enum';
+
+jest.setTimeout(10000);
```
christineyost, line 25: This test failed in CI but passed locally. The error message made it look like the setup did not complete before the default timeout of 5s. The test passed when I increased this timeout to 10s. Open to other solutions here though!

migrations/Migration20210802231612.ts
```diff
@@ -0,0 +1,55 @@
...
+      where counter.api_id = api.id
+    `);
+
+    // drop unnecessary api_id column from counter table
+    this.addSql('alter table "counter" drop column "api_id"');
+
+    this.addSql('alter table "api" alter column "metadata_counter_id" set not null;');
+    this.addSql('alter table "api" add constraint "api_metadata_counter_id_unique" unique ("metadata_counter_id");');
+
+    this.addSql('alter table "api_version" add column "internal_only" bool, add column "status" varchar(255);');
```
michaelkimball, line 30: Should we use an enum for `status`?
https://www.postgresql.org/docs/9.1/datatype-enum.html
I know we probably have limited types based on
https://github.com/department-of-veterans-affairs/lighthouse-di-api-styleguide/blob/07e3aa73bae836afc3f0f3a21d0f470def1cc07a/docs/long/versioning.md#api-lifecycle

src/api/version/apiVersion.schema.ts
```diff
@@ -12,6 +12,8 @@ export const apiVersionMutateSchema = Joi.object()
     securityType: Joi.string()
       .required()
       .allow(...Object.keys(SecurityTypeEnum)),
+    internalOnly: Joi.boolean().required(),
+    status: Joi.string().required(),
```
michaelkimball, line 16: enums again https://github.com/department-of-veterans-affairs/lighthouse-di-api-styleguide/blob/07e3aa73bae836afc3f0f3a21d0f470def1cc07a/docs/long/versioning.md#api-lifecycle

src/status/status.enum.ts
```diff
@@ -0,0 +1,8 @@
+enum StatusEnum {
```
michaelkimball, line 1: nice

jest.config.js
michaelkimball: Could we ignore `.model` files?

migrations/Migration20210802231612.ts
christineyost: Yes, we can do that! Today the only values I've seen for status are 'Previous Version' and 'Current Version'. Should we just support these two or should we support the ones mentioned in that style guide?

jest.config.js
christineyost: Yes we could, but it might be nice to ensure that at least the update methods are tested right? Or do you think the models are tested enough through testing the services?

migrations/Migration20210802231612.ts
michaelkimball: We should probably support the breadth of the styleguide

jest.config.js
michaelkimball: I think it's impossible to fully test the models as it Mikro-ORM hides a lot of it's implementation and we don't have direct access to it's methods. I think as long as we have coverage with the integration tests, we should be fine ignoring the models. That's just my 2 cents though, so feel free to ignore

christineyost: Sounds good to me! I think that's fair, we'll get way more out of the integration tests

src/api/version/apiVersion.schema.ts
```diff
@@ -12,6 +13,10 @@ export const apiVersionMutateSchema = Joi.object()
     securityType: Joi.string()
       .required()
       .allow(...Object.keys(SecurityTypeEnum)),
+    internalOnly: Joi.boolean().required(),
+    status: Joi.string()
+      .required()
+      .allow(...Object.keys(APIVersionStatusEnum)),
```
michaelkimball, line 19: Any reason for using `allow` over `valid`?
https://joi.dev/api/?v=17.4.2#anyvalidvalues---aliases-equal

christineyost: Got `allow` from above on the security type property, but `valid` does seem better here since it creates an exclusive list valid values

michaelkimball (approved):
✨ 🚀 ✨

PR #100, opened 2021-08-23
Add endpoints to create and get Metadata records

christineyost:
This PR is for API-8186. It adds endpoints that will allow users to create and view metadata records for APIs. To support this, the MetadataController and MetadataService were created. This PR also moves the `metadata` directory to be under the `api` directory since Metadata records belong to APIs.


src/api/metadata/metadata.controller.ts
```diff
@@ -0,0 +1,122 @@
...
+    } catch (err) {
+      next(err);
+    }
+  };
+
+  getMetadataByIteration: RequestHandler = (request, response) => {
+    response.json(request.metadata.content);
+  };
+
+  createMetadata: RequestHandler = async (request, response, next) => {
```
dsteeve, line 108: I would like to discuss this before approving.  By having to post to the /metadata endpoint can leave the data in an inconsistent state if the caller didn't know they had to post to this endpoint when creating a new version.  It might be better to not implement a POST and just have the apiVersion create and delete endpoints update the metadata.  

Not even sure we would need to version metadata then.  Whatever is out there is the latest and that should be ok? 

but I might be missing something.

dsteeve (changes requested):

PR #101, opened 2021-08-27
Add an endpoint to generate metadata for a given API

christineyost:
This PR adds a GET endpoint that will generate metadata for a given API. I ran into issues with storing metadata alongside the rest of the data because we wanted to store a new metadata version with each `APIVersion` create/update/delete, but then we need to worry about data transactions and potentially rollbacks since multiple writes to the database would happen in these operations. Generating metadata on the fly instead is not expensive and a much simpler solution. In the `upload` method we can also generate the metadata, then upload it to S3. If everyone agrees with this approach we can remove the metadata table with a later ticket.

Other notable changes:
- Move metadata directory under the api directory since metadata is related to an api specifically
- Change `getAPI` method in `APIService` to `doesAPIExist` since that's all this method was being used for


test/api/metadata/metadata.model.test.ts
```diff
@@ -11,7 +12,7 @@ describe('MetadataModel', () => {
         {
           version: '1.0.0',
           internal_only: true,
-          status: 'Current Version',
+          status: APIVersionStatusEnum.LIVE,
```
dsteeve, line 15: I know this wasn't part of this change, but we should probably get another ticket to refactor LIVE to CURRENT before this goes too much farther.   Since both a deprecated and a current version could be live at the same time, this seems named incorrectly.

dsteeve (approved):
approving, with some comments/suggestions.

dsteeve (approved):
thanks for making those changes!

PR #102, opened 2021-08-30
Add endpoint for uploading metadata to S3

christineyost:
This PR is for [API-8187](https://vajira.max.gov/browse/API-8187). It adds an endpoint that will generate metadata for a given API and upload the generated metadata to S3.


src/api/metadata/metadata.controller.ts
```diff
@@ -47,6 +61,17 @@ class MetadataController {
       next(err);
     }
   };
+
+  uploadMetadata: RequestHandler = async (request, response, next) => {
+    try {
+      const metadata = await this.metadataService.generateMetadata(request.api);
```
dsteeve, line 67: what happens if the api is misspelled or not found? Should this check that metadata is actually generated and exists before doing the upload or the other option is to catch the NotFoundError and handle gracefully since that is something the caller could fix instead of getting a 500 error.

christineyost: The [APIResolverMiddleware](https://github.com/department-of-veterans-affairs/lighthouse-docserver/blob/main/src/middleware/apiResolver.middleware.ts) will throw an error if the API is not found. The MetadataController makes use of the API resolver [here](https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/102/files#diff-2bb04ec03f9ebe43ed2d97b459f9cf37de2f251cb3d3e40e5820f7f17d8a4b63R45), so we should always have an API at this point. We could have a problem though if there are no APIVersions for that API. Do you think we should do that check?

dsteeve: we discussed offline, but will put notes here. 
yes.  I think if the api isn't there there will be no metadata, which will be the same problem when no apiVersions are there.  either way, we can throw a validationError from the service and have the controller handle that with a 400 bad request.  we should try to avoid sending out the stack messages to the caller - as that is considered a security flaw - but that problem can be handled more wholistically with a different ticket.

christineyost: Started a ticket for more complete error handling re-work here: https://vajira.max.gov/browse/API-9663

dsteeve: thank you

dsteeve (approved):
looks good!

PR #103, opened 2021-08-31
API-9039 Allow upload of OAS without using Kong for transformation

michaelkimball:
[API-9039](https://vajira.max.gov/browse/API-9039)

## Description

Changed shape of gatewayConfig to all optional and added removePath, exposedRoute to accomodate gatewayless transform
Updated Joi schema to ensure that either [serviceId, routeId] or [exposedRoute, removePath] are provided﻿
Changed transformer to either gateway-based or config-based transformation
Changed oas.service to decide which transformation based on what values are available (defaulting to config)
Updated/Added tests


dsteeve (dismissed):
looks good!

dsteeve (approved):

PR #104, opened 2021-09-08
API-9038 add upload on refresh in lower envs

michaelkimball:
[API-9038](https://vajira.max.gov/browse/API-9038)

Add upload when in dev or staging
Add tests to validate


src/oas/oas.service.ts
```diff
@@ -239,10 +239,18 @@ class OASService {
           result,
           apiVersion,
         );
+        void this.uploadInLowerEnvs(result);
```
christineyost, line 242: Should the upload only happen for the fetchAllOAS operation or should an OAS created by the POST operation also be automatically uploaded in lower environments?

src/oas/oas.service.ts
```diff
@@ -239,10 +239,18 @@ class OASService {
           result,
           apiVersion,
         );
```
christineyost, line 241: Should we still get a slack notification when the OAS is automatically uploaded?

ediiotero (approved):
LGTM 🚀

src/oas/oas.service.ts
```diff
@@ -239,10 +239,18 @@ class OASService {
...
+        void this.uploadInLowerEnvs(result);
       }
     }
   }
 
+  private async uploadInLowerEnvs(oas: OASModel): Promise<void> {
+    const env = process.env.OAS_ENV as string;
+    if (env === 'dev' || env === 'staging') {
+      await this.uploadOAS(oas);
```
christineyost, line 250: `updateOASStatuses` is the method responsible for marking the current OAS as ACTIVE and others as INACTIVE. That either needs to be called here or refactored so that it's automatically called from `uploadOAS`

michaelkimball: valid

src/oas/oas.service.ts
michaelkimball: I think if you're calling the post method specifically, then it's a small burden to call the upload

src/oas/oas.service.ts
michaelkimball: Yes, I think we still need to be kept in the loop when an OAS changes

christineyost (approved):
Looks good to me! 🎉

PR #105, opened 2021-09-16
Update links to mermaid and use npx instead of direct path to mmdh

michaelkimball:


mheusgit (approved):
Nice to have updated docs

PR #107, opened 2021-09-24
Add endpoints to return a delta between two OASs

christineyost:
This PR is for [API-9225](https://vajira.max.gov/browse/API-9225). It adds two endpoints: `docserver/apis/{api-name}/versions/{versionNumber}/oas/latest/delta` and `docserver/apis/{api-name}/versions/{versionNumber}/oas/{oasIteration}/delta`. The first will return a delta between the latest OAS and the current ACTIVE one. The second will return a delta between the OAS with the given iteration and the current ACTIVE one. If there is no current ACTIVE OAS, the delta will be the latest/iteration OAS in its entirety. If there is no difference, the delta will be undefined. More information on these deltas can be found [here](https://github.com/benjamine/jsondiffpatch/blob/master/docs/deltas.md).

This PR also changes the order we pass the OASs to the differ because according to the documentation we previously had them backwards.


test/oas/oas.service.test.ts
```diff
@@ -121,7 +121,7 @@ describe('OASService', () => {
       service.fetchOAS = jest.fn().mockResolvedValue(oasModel);
       service.entityManager.findOne = jest.fn().mockResolvedValue(oasModel2);
       const diff =
-        '{\n  "info": {\n    "version": [\n      "0.0.0",\n      "old"\n    ]\n  }\n}';
+        '{\n  "info": {\n    "version": [\n      "old",\n      "0.0.0"\n    ]\n  }\n}';
```
dsteeve, line 124: will this fix affect the slack autocomparison result?  like is the resulting value being displayed now be the wrong one?

dsteeve (approved):
Looks great, I just had a couple questions I put in for discussion.

test/oas/oas.service.test.ts
christineyost: What is being displayed in Slack right now is backwards according to the [documentation](https://github.com/benjamine/jsondiffpatch/blob/master/docs/deltas.md). It specifies that when a value is modified the delta will be `[ oldValue, newValue ]` which is what Slack will display after this change goes in. Currently Slack is displaying `[ newValue, oldValue ]`. This problem also results in our Slack message displaying a new item as getting deleted and a deleted item being shown as new. This should all be corrected since we're switching the order the new and old items are sent to the differ [here](https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/107/files#diff-d6dd283f58b97f16af865bc07cb1b2c44f7e896a16de55c2f5dccf707a1ffefbR21)

dsteeve: awesome!  thanks for explaining.

src/differ/differ.service.ts
```diff
@@ -13,13 +13,13 @@ export class DifferService {
     });
   }
 
-  public diffOAS(newOAS: OAS, oldOAS?: OAS): string {
+  public diffOAS(newOAS: OAS, oldOAS?: OAS): Delta | undefined {
```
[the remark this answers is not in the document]
dsteeve: we learned in our testing today that it will return an empty object since Delta is an object.  so it returns {} when the json is parsed and there is no Delta.

PR #109, opened 2021-09-27
Api 9960/update readme data dictionary

mheusgit:


dsteeve (approved):
looks great!

dsteeve:
should the readme reference the confluence page, so if the data dictionary is changed it is changed in both places?

PR #111, opened 2021-10-07
Api 9040/responses

christineyost:
This PR is for API-9040. It updates OAS transform logic so that standard gateway 401, 403, and 429 responses are added to all operations for APIs secured with API keys.

I thought logic for transforming paths was getting a bit complex inside of `oas.service.ts` so I moved that logic to `oas.transformer.ts` because I thought that logic seems like it belongs there more than in the service. However when I did that I ended up with a bit of an explosion of the `oas.transformer.test.ts` file. Looking for some feedback on this. Maybe it would be better to move some of those path transformation helpers to a new class and test them separately then mock them out in `oas.transformer.test.ts`? Open to other ideas as well.


test/fixtures/oneRingOAS.ts
```diff
@@ -123,6 +123,56 @@ export const oneRingOASTransformed = {
           '200': {
             $ref: '#/components/responses/OneRingResponse',
           },
+          '401': {
```
dsteeve, line 126: instead of hardcoding these response multiple times, is it possible to just do a  + response401 +", " + response403 + "," + response429 ?  Then we only have to maintain in one place?  Or maybe create a testGatewayResponse const and then + that to the response?

I think it was a good move to refactor the transformation out of the service.

ediiotero (approved):
I like moving more of the transforming logic to oas.transformer. Interested in what the rest of the team thinks, but moving the path transformation helpers to a new class seems like a good idea. Would we want to do this work now or in a new ticket? I'm good with it either way

dsteeve (approved):
looks like a great improvement to break this out a bit!  I think the refactor captures our desired outcome.

PR #114, opened 2021-10-12
API-10196  Update non-gateway path transformation logic

mheusgit:
Moved normalizeLeadingAndTrailingSlash() to top of updatePathsElement() due it is being shared logic between the two approaches.


mheusgit:
Updated PR to include revised test cases.

PR #117, opened 2021-11-03
API-10199  Write an OAS for DocServer - Update to follow AWS routing

mheusgit:
Additional change since ECS is enforcing a "/docserver/*" URL policy against docserver requests.


christineyost (approved):
🎉

PR #118, opened 2021-11-03
API-10199  Write an OAS for DocServer - node & path fix

mheusgit:
Update to disable secondary NODE_ENV check & alter swagger path


christineyost (approved):
🚀

PR #119, opened 2021-11-04
Api 10858    docserver  update to use better practices for variables

mheusgit:
Changes in this PR:
- Newly added config.ts greatly reduced configuration sprawl throughout the app by centralizing configuration setup & default handling
- Added jestSetEnvVars.ts to manage Jest environment variable setup 
- Updated ReadMe concerning new expectations for .env values
- Removed hardcoding of variables throughout the app
- Added "startup:local" script to further centralize environment variables
- Added errors if docker-compose files are missing required variables
- Fixed bug connected to misspelled NODE_ENV


christineyost:
Looks like the build failure might be due to the `ECR_REGISTRY` variable in `build.sh`?

mheusgit:
> 
> 
> Looks like the build failure might be due to the `ECR_REGISTRY` variable in `build.sh`?

That is my understanding as well.  I expect the changes build.sh and test-database.yml will require additional environment variables to be added to Jenkins.

dsteeve (dismissed):
Assuming you can get the build issue figured out, all the other changes look good.  I ran it locally and it seemed fine.

mheusgit:
Jenkins fixes are in.  This ticket should be good to go.

src/config.ts
```diff
@@ -0,0 +1,96 @@
...
+      description: 'Production',
+    },
+  ],
+};
+
+const env = process.env.OAS_ENV || 'unknown';
+const config = {
+  ENV_NAME: env,
+  NODE_ENV: process.env.NODE_ENV || 'unknown',
+  ENV_METADATA: ENV_METADATAS[env],
```
christineyost, line 47: These are also defined in the refresh function below. Do they need to be set here since that method is called at the end of this file?

test/oas/oas.service.integration.ts
```diff
@@ -59,21 +60,22 @@ describe('OASService', () => {
...
+      region: cfg.AWS_REGION,
     });
     notificationService = ({
       sendUpdatedOASMessage: jest.fn(),
     } as unknown) as NotificationService;
     const gatewayService = new GatewayService({
-      username: 'gimli',
-      password: 'andMyAxe',
+      username: 'fake-username',
+      password: 'fake-password',
```
christineyost, line 78: Should these come from `cfg`?

christineyost (approved):
Just had a couple small questions, but overall looks good to me!

src/config.ts
mheusgit: If I remember correctly, I left the sets in the first location so Typescript wouldn't designate them a 'any' type

PR #120, opened 2021-11-05
API-10199  Write an OAS for DocServer - build fix

mheusgit:
Update to insure Docker build includes required files


christineyost (approved):
🚀

PR #121, opened 2021-11-05
Update refresh endpoint

ediiotero:
Story: [API-11189](https://vajira.max.gov/browse/API-11189)

Remove the check for lower env. It will refresh and upload in all env


mheusgit (dismissed):
Changes look good

christineyost:
Looks like the OAS controller integration test just needs a longer timeout set?

src/oas/oas.service.ts
```diff
@@ -183,19 +183,12 @@ class OASService {
           result,
           apiVersion,
         );
-        await this.uploadInLowerEnvs(result);
+        await this.uploadOAS(result);
+        await this.updateOASStatuses(result.apiVersion);
```
dsteeve, line 187: just fyi... I don't love this that we are doing updates in a method that says "fetch",  but that can be addressed in a separate ticket. This change isn't making it any worse than before.
see: https://vajira.max.gov/browse/API-11191

dsteeve (approved):

dsteeve (approved):

PR #124, opened 2021-11-23
Make isModelDifferent smarter

ediiotero:
Story: [API-11192](https://vajira.max.gov/browse/API-11192)

Moved transform to fetchOAS so we are hashing the transformed spec. This should reduce the number of saved iterations.


christineyost (commented):
This change looks good to me overall.

Wanted to note that this will cause one additional iteration for every APIVersion in which the transformed OASs may be identical since the method for creating the hash changed, but I think that's ok because after that one we will only get new iterations when the transformed OAS is different.

I thought it might be nice to add a test where the original spec changes, but the transformed OAS is the same and assert that a new iteration is not created after the change

peyton-lucas:
Changes look good to me.

christineyost (approved):
Looks good to me 🎉

PR #125, opened 2021-11-29
Api 10361  doc server mikro orm migrations are out of sync with annotations

mheusgit:
PR includes the three fixes from previously reported issue


christineyost (approved):
Looks good to me!

PR #126, opened 2021-12-01
Upgrade Docserver Dependencies

ediiotero:
Story: [API-10728](https://vajira.max.gov/browse/API-10728)

This PR upgrades node to 16.13.1 and makes updates to the tool versioning files. I also upgrades the node image being used in the Dockerfile.

- Tested a few endpoints everything seems to be working
- Lockfile is now at v2
- Also added changes from [API-11519](https://vajira.max.gov/browse/API-11519)


peyton-lucas:
Looks good to me!

dsteeve (approved):
lgtm!

dsteeve (approved):
I figured out what was going on, played with it some more.  I think we are good to go.

PR #129, opened 2021-12-21
API-11689: only return the oas if the status is active or review

dsteeve:
only return the oas if the status is active or review -- otherwise too much memory can be used causing docserver issues.


christineyost (approved):
Looks good to me!

PR #131, opened 2022-01-07
API-12336: Removed logging filter and added additional logging as requests come in

stevelong00:
- Removed filter that was preventing health check endpoint hits from being logged
- Added a 2nd logger middleware to log requests as they come in, even if they subsequently error out


mheusgit (approved):
Looks good

PR #133, opened 2022-01-13
Api 12222  doc server large oas memory leak   uncaught error fixes

mheusgit:
First PR in a set of fixes to resolve DocServer repeating shutdown issue


src/app.ts
```diff
@@ -123,3 +125,21 @@ function setupRequestOasGenerator(): void {
...
 }
+
+//Node 15 changed uncaught exception/rejection behavior to automatically stop node
+//  Method overrides default behavior and provides helpful logging instead
+function setupUncaughtErrorFallbacks(): void {
+  process.on('unhandledRejection', (reason, promise) => {
+    logger.info('----- Unhandled Rejection at -----');
+    logger.info(promise);
+    logger.info('----- Reason');
+    logger.info(reason);
```
christineyost, line 136: Should we log these and other messages in this PR at the error level instead of info?

src/oas/oas.service.ts
```diff
@@ -203,7 +203,7 @@ class OASService {
           delta = JSON.stringify(delta, null, 2);
         }
         logger.info(`Sending Slack notification with delta: ${delta}`);
-        void this.notificationService.sendUpdatedOASMessage(
+        await this.notificationService.sendUpdatedOASMessage(
```
christineyost, line 206: Along with this change can we move sending the slack message to the end of this function so we can hopefully get the data all set up correctly before we potentially throw an error when sending the slack message?

src/app.ts
mheusgit: We have a ticket to rework how the logging is done across the repo.  I want this to be part of the debug logging that is planned.

src/oas/oas.service.ts
mheusgit: I'll move it for the time being.  The method needs to be redesigned so that messages are not lost if a failure happens and triggers a reattempt later.

src/app.ts
dsteeve: I think this should be .error for now instead of .info.  That way we can better search in the meantime.

src/oas/oas.service.ts
dsteeve: that is a good idea.  If this sendupdatedOASMessage is moved to after the updateOASStatuses, that will be better for now.

mheusgit:
Requested changes are in place.  Ticket should be good to go.

mheusgit:
Critical fix added to PR for file apiVersion.service.ts

src/notification/notification.service.ts
```diff
@@ -83,10 +84,19 @@ export default class SlackService implements NotificationService {
       // will accept and log that field.
       const { response } = err as AxiosError;
       if (response) {
+        logger.info(`SlackService: Response error`);
```
dsteeve, line 87: I think these should be logged as error since it is reporting an error.

dsteeve (commented):
overall looks good Marc.  I just had one question and one minor change.

src/app.ts
```diff
@@ -123,3 +125,21 @@ function setupRequestOasGenerator(): void {
     expressOasGenerator.handleRequests();
   }
 }
+
+//Node 15 changed uncaught exception/rejection behavior to automatically stop node
+//  Method overrides default behavior and provides helpful logging instead
```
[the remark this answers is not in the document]
mheusgit: This will keep the server up.

mheusgit:
Requested changes are in place

dsteeve (approved):
I think this looks good! thanks for making that one change.

PR #134, opened 2022-01-19
API-12280-docserver: turn OASGenerator back on

stevelong00:
API-12280-docserver: turn OASGenerator back on

OAS generator in back to being always-on in dev environment


mheusgit:
The change is correct but Jenkins is throwing an error

mheusgit (approved):
A reattempt resolved the build error.  Everything else is good.

PR #135, opened 2022-02-02
API-13118 DocServer - add request headers in the logger

stevelong00:
Added host, referer, and user-agent to morgan logs.


mheusgit (approved):
Nice to see this in place

PR #142, opened 2022-02-18
API-12279/ update getting all OASs by status

ediiotero:
Story: [API-12279](https://vajira.max.gov/browse/API-12279)

This PR addresses some of the problems found during [API-11689](https://vajira.max.gov/browse/API-11689) and link to that PR [here.](https://github.com/department-of-veterans-affairs/lighthouse-docserver/commit/1cfcfa81d597c502da816fe86b2328a88de2a812)

Ran into a problem with the upload endpoint. After the changes I made (similar to the change in API-11689) I noticed I could upload an inactive OAS but the status would not change to active because is was not included in the list when getting all the OASs by status. I fixed this by adding an optional parameter to updateOASStatuses and passing active, inactive, and review when it is called in the upload endpoint. I'm open to suggestions as I know this should probably get all if we need to upload an error status and that might be a lot to get back but since its only when we upload I thought is would be ok.


mheusgit:
Updated PR title to aid JIRA hooks

ediiotero:
took another look at the ticket and forgot to make a change so we could not upload an OAS with an error status. Will write a ticket to add an endpoint to change an OAS status

src/oas/oas.controller.ts
```diff
@@ -201,10 +208,18 @@ class OASController {
     try {
       const { oas, apiVersion } = request;
 
-      await this.oasService.uploadOAS(oas);
-      await this.oasService.updateOASStatuses(apiVersion, oas.iteration);
+      if (oas.status === StatusEnum.ERROR) {
+        response.sendStatus(400);
```
mheusgit, line 212: How will calls to uploadLatestOAS() function if the latest OAS turns out to be in a error state?  Wont a request to '/latest/upload' return with a 400 error but still be valid?

Maybe uploadLatestOAS() should be for the latest "valid" OAS

src/oas/oas.controller.ts
```diff
@@ -150,12 +151,18 @@ class OASController {
     response.json(request.oas.transformedSpec);
   };
 
-  //TODO:  this needs a filter as returning all is too many, so thus this doesnt return all (API-11689)
   listOASForAPIVersion: RequestHandler = async (request, response, next) => {
```
christineyost, line 154: Does the status default (ACTIVE and REVIEW) make sense for this endpoint?

src/oas/oas.controller.ts
mheusgit: Yeah.  That would be the right approach.

test/oas/oas.service.integration.ts
```diff
@@ -357,7 +361,7 @@ describe('OASService', () => {
     });
 
     it('updates status as expected on roll back', async () => {
-      await service.updateOASStatus(firstFetchedOAS, StatusEnum.INACTIVE);
+      await service.updateOASStatus(firstFetchedOAS, StatusEnum.REVIEW);
```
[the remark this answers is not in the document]
ediiotero: Makes sense, I think I missed something here. I'll fix it.

src/oas/oas.controller.ts
ediiotero: Not sure. It's probably a better idea to return everything or maybe include INACTIVE in the default. I thought we were trying to reduce the size of the data coming back because of all the iterations. Maybe I can return a range of iterations?

dsteeve: right, we can't include all inactive statuses.  It is too much data. 
we might need to sit down and think this through and work through all the use cases.

src/oas/oas.controller.ts
dsteeve: correct, otherwise you could never get past this iteration in error - since a delete option isn't there yet.  You would have to make some change and do a refresh to get past it.

src/oas/oas.service.ts
```diff
@@ -176,6 +176,35 @@ class OASService {
...
+        throw new Error(
+          `Expected 1 OAS for ${apiVersion.api.name} v${apiVersion.majorVersion} with iteration ${iteration}, but found ${apiVersion.openAPISpecs.length}`,
+        );
+      }
+
+      const oas = apiVersion.openAPISpecs[0];
+      const nextIteration = oas.iteration - 1;
+
+      return oas.status === StatusEnum.ERROR
+        ? this.getNonErrorOASForAPIVersion(apiVersion, nextIteration)
```
mheusgit, line 198: This works for now.  If we run into performance issues, will be rewritten so that a query gets the first valid iteration.

src/oas/oas.service.ts
```diff
@@ -134,16 +134,14 @@ class OASService {
...
-  public async getAllOASForAPIVersion(
+  public async getOASIterationsByStatuses(
     apiVersion: APIVersionModel,
+    status: StatusEnum[] = [StatusEnum.ACTIVE, StatusEnum.REVIEW],
   ): Promise<OASModel[]> {
     try {
       await apiVersion.openAPISpecs.init({
         orderBy: { iteration: 'desc' },
-        where: { status: { $in: [StatusEnum.ACTIVE, StatusEnum.REVIEW] } },
+        where: { status: { $in: status } },
```
mheusgit, line 144: Good improvement

dsteeve (approved):
looks good! shouldl fix the "ERROR" issue we talked about.  Thanks!

src/oas/oas.service.ts
```diff
@@ -178,29 +183,25 @@ class OASService {
...
     try {
       await apiVersion.openAPISpecs.init({
-        where: { iteration },
+        where: { status: { $ne: StatusEnum.ERROR } },
+        orderBy: { iteration: 'desc' },
       });
+      const oasList = apiVersion.openAPISpecs.getItems();
 
-      if (apiVersion.openAPISpecs.length !== 1) {
+      if (oasList.length < 1) {
```
mheusgit, line 194: Good change.  These Array().length !== number conditions are not ideal

mheusgit (approved):
Looks good

PR #143, opened 2022-02-22
Remove console.log statements from Docserver

christineyost:
This PR is for [API-13328](https://vajira.max.gov/browse/API-13328). It removes two `console.log` statements from Docserver. These statements were replaced with `logger.error` since in both cases an error was being logged to the console.

This PR also adds the eslint `no-console` to our eslint configuration. With this rule set to `error` when a `console.log` statement is added to Docserver linting will fail. This rule can be turned off if necessary by adding a comment above the line. More info here: https://eslint.org/docs/rules/no-console


mheusgit (approved):
Looks good.

dsteeve (approved):
looks good!

PR #144, opened 2022-02-23
API-13607/  Add requestUUID to the logs

mheusgit:
Change required
- Adding express-http-context & uuid modules
- Updated both logger middleware
- Added uuid formatter to winston logging


mheusgit:
Running build again fixed the Jenkins error

christineyost (commented):
Docserver has an established pattern for attaching context to requests. Custom properties are defined in @types/express/index.d.ts. There is a corresponding middleware for each property which sets the custom property on the request (src/middleware/apiResolver.middleware.ts to set the API for example).

I think if we want to use express-http-context instead that's ok, but I think we should also create a ticket to move these custom request fields to use this new pattern as well so we don't have two different ways of tracking custom fields on requests.

I do like that with the established pattern we can use strong typing from Typescript and wouldn't need the eslint disable statements when getting custom request properties.

mheusgit:
At first glace the two approaches look interchangeable but they truly are not.  The resolver middlewares approach is limited to being accessible in locations where the request is passed such as RequestHandlers.  At that point we have to manually pass additional arguments to called methods.  The express-http-context module exposes the context everywhere without passing arguments and supports calls outside a request such as server startup.

I will try to get ride of the eslint statements.

mheusgit:
The eslint disable were removed

src/middleware/loggerRequest.middleware.ts
```diff
@@ -25,7 +26,10 @@ function loggerRequestMiddleware(
   req: Request,
   res: Response,
 ): string {
+  const uuid: string = getContext('uuid') as string;
```
christineyost, line 29: 🎉

christineyost (approved):
Looks good to me!

PR #146, opened 2022-03-01
Stream request and response logs to Winston logger

christineyost:
This PR is for [API-13438](https://vajira.max.gov/browse/API-13438). It updates request and response logging so that the messages are streamed to the Winston logger. Logs for the healthcheck endpoint are logged at the verbose level and other requests/responses are logged at the info level. I removed the uuid field from the morgan logs since this field is now added when these messages are streamed to the Winston logger.


mheusgit (approved):
Looks good

dsteeve (approved):
lgtm 👍

stevelong00:
Looks good!

PR #148, opened 2022-03-03
API-13187/  Provider Dashboard: Expose Updated At data

mheusgit:
- Change includes "updatedAt" data to now be part of the API version responses


stevelong00 (approved):
Looks like it should work

PR #150, opened 2022-03-08
API-13330  DocServer: log when communicating to 3rd parties

mheusgit:
- Added request/response logging to 3rd party request


dsteeve (approved):
this looks great!

PR #151, opened 2022-03-10
Api 12399  doc server  handling resources that don't exist

mheusgit:
- Updated Express error handling middleware to output response bodies matching IETF suggested approach.  Change removed stack trace from response bodies. Server error to now include status codes, description, type URLs, and multiple reason support.
- Revised Error handling connected to API version lookup failures to follow standard used elsewhere
- Error response bodies across environments are now consistent
- Migrated API & API version lookup to new error creation approach
- Updated unit test accordingly


src/utils/serverError.ts
```diff
@@ -1,9 +1,48 @@
...
     this.stack = err.stack;
+    this.statusCode = statusCode;
+    this.description = description;
+    this.detailTitle = detailTitle;
+    this.details = details;
+    this.type = type;
+  }
+}
+
+export class DataResolveError extends ServerError {
```
christineyost, line 34: Maybe we should rename this file to `errors.ts` or perhaps define each error type in a separate file, especially if we may add more error types in the future?

test/api/api.service.test.ts
```diff
@@ -99,7 +99,7 @@ describe('APIService', () => {
     } catch (err) {
       if (err instanceof ServerError) {
```
christineyost, line 100: can we confirm that this err is actually an instance of DataResolveError?

test/api/version/apiVersion.service.test.ts
```diff
@@ -183,7 +183,9 @@ describe('APIVersionService', () => {
       } catch (err) {
         if (err instanceof ServerError) {
```
christineyost, line 184: can we check if this err is a DataResolveError?

christineyost (commented):
Looks good! I like this pattern of introducing new error types!

We have some similar logic for OAS iterations as well where we're throwing errors if a certain iteration doesn't exist, like in oas.service getOASForAPIVersion. We could possibly also use the DataResolveError there.

src/utils/serverError.ts
mheusgit: Yeah, it needs a rename

mheusgit:
["We could possibly also use the DataResolveError there."](https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/151#pullrequestreview-907816509)
Ticket scope is limited to just the two scenarios but more will be added later including OAS not found.

test/api/api.service.test.ts
mheusgit: Sure

test/api/version/apiVersion.service.test.ts
mheusgit: Yep

src/api/version/apiVersion.service.ts
```diff
@@ -96,17 +96,24 @@ class APIVersionService {
...
   ): Promise<APIVersionModel> {
-    const version = await this.getAPIVersion(api, majorVersion);
+    try {
+      const version = await this.getAPIVersion(api, majorVersion);
 
-    if (version === null) {
-      throw new ServerError(
-        new Error(
+      if (version === null) {
+        throw new Error(
```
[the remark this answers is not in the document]
mheusgit: It can but I am not a fan.  await this.getAPIVersion(api, majorVersion) can throw a different type of exception and it causes the code to mismatch against how exceptions are handled elsewhere.  Below is what you would end up with

`throw new DataResolveError(new Error(
          `APIVersionModel not found ({ name: ${api.name}, majorVersion: ${majorVersion} })`,
        ), [
        {
          name: 'apiVersion',
          reason: `API version '${majorVersion}' not found for API '${api.name}'`,
        },
      ]);`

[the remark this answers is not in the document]
mheusgit: I'll see if I can manage a overload of the constructor.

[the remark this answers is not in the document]
mheusgit: Updated the way ServerErrors are done to make the Error wrapping optional

christineyost (approved):
Tested locally - logging and errors returned to the client look good to me!

dsteeve (approved):
I brought this branch down and tested out a few things I had questions on, and it looks like a viable pattern we can use.  I think this is a huge leap forward in handling different types of errors and giving better responses.  
Looks good! 👍

mheusgit:
Branch contained conflicts that were not picked up on by Github.  A new PR has been created to fix.
https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/153

PR #152, opened 2022-03-15
Api 14212  decision reviews v2 oas is not getting refreshed

mheusgit:
- Fixed issues with isModelDifferent() logic.  Will only consider the previous iteration during compare and provides relevant logging.
- Resolved logic in refreshOAS() that could trigger unrelated API refreshes to fail if a single error was encountered for one API
- Added failure Slack notification messages
- Moved some logic out of refreshOAS() to a better named refreshAllOAS()


src/oas/oas.service.ts
```diff
@@ -284,12 +305,33 @@ class OASService {
...
+    const latestModel = await this.entityManager.findOne(
+      OASModel,
+      {
+        apiVersion: apiVersion,
+      },
+      { orderBy: { iteration: 'desc' } },
+    );
+
+    if (latestModel) {
+      if (latestModel.hash !== hash) isDifferent = true;
```
dsteeve, line 322: personal opinion, but I find it easier to read if {} are added - and less future mistakes are made when changing the code.

src/oas/oas.service.ts
```diff
@@ -178,46 +178,67 @@ class OASService {
...
+        logger.info(`Refreshing api: ${apiVersionStr} `);
+
+        await this.refreshOAS(apiVersion);
+      } catch (err) {
+        const failureMsg = `Failed to refresh API: ${apiVersionStr}`;
+        logger.error(failureMsg);
+        logger.error(err);
+
+        logger.error(`Sending Slack notification for failure`);
+        await this.notificationService.sendFailureMessage(
```
dsteeve, line 196: I like what you are trying to do here. However, if slack has a problem, this will still fall out of the loop and not process anything more. 

It might be better to set an error flag of "sendFailureSlack" or something and then send the notification in a different try/catch after checking that flag.  Typically isn't good to do complex logic in a catch.

mheusgit: I will wrap it one another try/catch.  Shouldn't be too complex.

src/oas/oas.service.ts
christineyost: or maybe make this one if statement with an `&&`?

mheusgit: Added the brackets

src/oas/oas.service.ts
mheusgit: Change is in.  We could respond with a different status code but I don't think anything is depending on that at the moment.

src/oas/oas.service.ts
```diff
@@ -206,49 +206,74 @@ class OASService {
...
+        logger.error(err);
+
+        try {
+          logger.error(`Sending Slack notification for failure`);
+          await this.notificationService.sendFailureMessage(
+            failureMsg,
+            err as Error,
+          );
+        } catch (err) {
+          logger.error(`Slack notification for failure could not be sent`);
```
dsteeve, line 230: the apiVersionStr should probably be part of this log message, but the uuid should tie them together as well.

dsteeve (approved):
small potential improvement, but otherwise looks good.

PR #154, opened 2022-03-17
Update OAS Transformer for APIs secured by API key

christineyost:
This PR is for API-14234. It updates the OAS Transformer for APIs secured by API key so that it first checks existing security schemes for one of type `apikey`. If one like this exists, the transformer will not create a new one.

Also added a couple simple OAS fixtures to support unit testing.


mheusgit (approved):
Looks good

PR #155, opened 2022-03-18
API-14361/  benefits intake v1 &v2 refresh failures

mheusgit:
Changes include:
- Bug fix for OAS difference detection logic.  New OAS will be created only if jsondiffpatch detects a none empty change.  JSON sequence changes will no longer be used as a reason for new OAS generation.
- Refactor of refreshOAS & fetchOAS logic in oas.service.ts.  The sequence of events have been rearranged so that new models will be created only after a none empty delta is encountered.  Moved delta logic under resolveDelta()
- Renamed OASController createOAS() to refreshOAS()
- Number of test changes connected to refactor and renaming


src/oas/oas.service.ts
```diff
@@ -233,45 +209,59 @@ class OASService {
...
+      const currentData: OAScreate = {
+        spec: currentOAS,
+        apiVersion,
+        hash: currentHash,
+        transformedSpec,
+      };
+
+      const delta = await this.resolveDelta(currentData);
+
+      if (delta && delta !== '') {
```
dsteeve, line 239: future state, it might be better to have an 'else' with this, and if delta is undefined or empty, then log that here instead of where it is currently being logged.

PR #156, opened 2022-03-21
API-14361  Benefits-intake v1 &v2 refresh failures - Text length fix

mheusgit:
- Fix for invalid_blocks errors that Slack has been returning due to size constraints


peyton-lucas (approved):
Looks good to me!

PR #157, opened 2022-03-23
API-14249/  docserver: handle the updating of an oas iteration status

mheusgit:
- Updated OAS update endpoint to be more flexible for changing OAS status
- Added new type of error 'ValidationError'
- Reworked a number of previously existing test cases to work with new endpoint


stevelong00 (approved):
Looks good!

PR #158, opened 2022-03-25
Update delta generation to use transformed spec instead of original

christineyost:
This PR is for [API-14749](https://vajira.max.gov/browse/API-14749). It updates the JSON delta generation logic to use the transformed spec to generate the delta instead of the original. With this change if transformation logic changes we will get a new OAS iteration the next time refreshAll runs even if the original OAS does not change. This is how refreshAll worked previously.


mheusgit (approved):
Change look fine.  I expect this will cause a number of OAS to get new iterations all at once

peyton-lucas (commented):
Looks good to me!

src/oas/oas.service.ts
```diff
@@ -428,7 +428,10 @@ class OASService {
 
     // Make delta variable mutable
     let delta: Delta | undefined | string;
-    delta = this.differService.diffOAS(currentData.spec, latestModel?.spec);
+    delta = this.differService.diffOAS(
```
dsteeve, line 431: will this actually get called?  OR will Line 426 short circuit since the model hash is the same?

christineyost: Line 426 will not short circuit because isModelDifferent compares hashes which are generated using a combination of the API name, version number, and the transformed OAS on line 229. As long as the transformed OASs are different, the isModelDifferent function will report that there is a difference.

dsteeve: sounds good. thank you for verifying

dsteeve (approved):
👍

PR #159, opened 2022-03-25
Api 13610  doc server  rest api to change logging levels

mheusgit:
- Added new log endpoint
- Change required new controller, service, and corresponding test


src/admin/admin.interface.ts
```diff
@@ -0,0 +1,3 @@
+export interface Logs {
+  logLevel: string;
```
christineyost, line 2: What happens if an invalid log level is sent in? Should this be an enum?

christineyost (commented):
Looks great to me! Just wondering what the behavior is if an invalid logger level is passed in

src/admin/admin.interface.ts
mheusgit: If someone enters a invalid log level, winston will switch to that level of logging regardless.  If someone enters gibberish as a log level the normal outcome will be no logs being displayed until a meaningful log level is used.

I thought of using a enum to force it but didn't since winston permits it.

christineyost: Ok, that works for me

christineyost (approved):

PR #162, opened 2022-04-15
Api 15090  docserver  delete inactive iterations

mheusgit:
- Change added a new endpoint that can perform a bulk cleanup of inactive and old OAS iterations


src/oas/oas.controller.ts
```diff
@@ -243,6 +249,29 @@ class OASController {
     }
   };
 
+  deleteOAS: RequestHandler = async (request, response, next) => {
+    const { apiVersion } = request;
+    const {
+      oasStatus,
+      olderThan = 0,
```
christineyost, line 256: Is olderThan required or not? Looks like it's required in `OASDeleteQuery` but not in `oasIterationDeleteSchema`?

src/oas/oas.service.ts
```diff
@@ -317,6 +318,35 @@ class OASService {
     await this.entityManager.persistAndFlush(oas);
   }
 
+  public async deleteOAS(
+    apiVersion: APIVersionModel,
+    oasStatus: StatusEnum,
+    expireDate: Date,
+  ): Promise<number> {
+    if (oasStatus !== 'inactive') {
```
christineyost, line 326:
```suggestion
    if (oasStatus !== StatusEnum.INACTIVE) {
```

src/oas/oas.controller.ts
```diff
@@ -243,6 +249,29 @@ class OASController {
...
+    } = (request.query as unknown) as OASDeleteQuery;
+
+    try {
+      const expireDate: Date = new Date();
+      expireDate.setDate(expireDate.getDate() - olderThan);
```
christineyost, line 261: The ticket mentions deleting OASs that are older than x days from the update_at or created_at times, but maybe it's better to go from today instead?

christineyost (commented):
Looks good to me! Just had a couple questions

src/oas/oas.controller.ts
mheusgit: It defaults to the current time if a day is not provided

src/oas/oas.service.ts
mheusgit: I'll update that

src/oas/oas.controller.ts
mheusgit: The plan is to setup a cronjob to run this on a periodic basis at some point.  We probably want to clear out OAS after 90 days.

src/oas/oas.service.ts
mheusgit: This change was added

src/oas/oas.controller.ts
dsteeve: I think I see Christine's point.  Maybe default older-than to 1 instead of 0.  Older than 24 hours basically.

test/oas/oas.controller.integration.ts
```diff
@@ -244,5 +244,36 @@ describe('OAS Controller', () => {
         expect(response.status).toBe(401);
       });
     });
+
+    describe('DELETE /docserver/apis/:apiName/versions/:majorVersion/oas', () => {
+      it('returns with 201', async () => {
+        //Full scope of this test should include OAS upload requests to support deletions but it is waiting on S3 fix
```
dsteeve, line 250: I don't think I understand this comment.  A delete of an OAS should never be an active one, so the S3 value shouldn't be affected?

dsteeve (commented):
looks good, I just had a question

dsteeve (dismissed):
After further reflection, I am good with leaving this deleting all iterations that are inactive with default be 0.  The caller can always use 1 to get the desired behavior.

mheusgit:
Failing checks are originally from another PR and should have been caught beforehand.  Seems to be due to Jenkins builds not automatically getting rerun after another PR is merged.  Something to look into.

christineyost (approved):
Looks good to me!

test/oas/oas.controller.integration.ts
christineyost: I think the problem here is in the `beforeEach` an OAS is added with a POST. That will add the OAS in review status. To get some OASs in inactive status we need to upload one.

Now that that fix is in though can this be updated?

mheusgit: Yes.  With that fix in I'll update the test

christineyost (approved):
🎉

PR #163, opened 2022-04-18
Add refresh to apiVersions

ediiotero:
Story: [API-14382](https://vajira.max.gov/browse/API-14382)

- Added new column to API versions table "refresh" not null default true.
- Refresh can be updated with the update endpoint and can be set when adding an API version.
- Refresh endpoint will only refresh APIs with refresh set to true.
- Updated readme


christineyost (commented):
Looks good to me!!

I'm curious what others think of requiring the `refresh` property to be sent in to the POST endpoint for APIVersion creation. It seems like we'd most of the time want this to be true so I'm wondering if we should default it to true if it's not provided. If we do that we wouldn't have to require this field for the POST endpoint.

However, requiring it also makes sense to me because I think we should require it for the PUT endpoint for updating an APIVersion. Requiring the field here makes sense to me because we require every property for update endpoints throughout the code. Then if we also require this field for the POST endpoint the request body for these endpoints would be the same, which seems like it would also be nice

I don't feel strongly either way, but wondering if others have opinions?

mheusgit:
Everything looks good to me as well.  

Its pretty wild how this change required 27 files to be changed though.  This ticket is going to have allot of conflicts with [API-14251](https://vajira.max.gov/browse/API-14251)

mheusgit:
> 

I'd prefer most fields to be optional.  As the number of features grow it will become an annoyance to test with a 'required' bias or revert them to optional down the road.

Defaulting to "true" and allowing someone to set to "false" later will be the normal use case.

christineyost (approved):
🚀

PR #166, opened 2022-04-20
Transform operationId during OAS transformation

christineyost:
This PR is for [API-14967](https://vajira.max.gov/browse/API-14967). It updates OAS transformation so that it includes the transformation of operationIds. If an operation did not previously have an operationId one will be added, and if it did have an operationId the value will be overwritten. The operationId will follow the pattern `<HTTP_METHOD>:<PATH>`, for example `GET:/facilities/{id}`

This PR includes significant refactoring of the `OASPathsTranformer` class. I noticed this class was iterating through each path item a couple times so I updated it to only iterate through these once, transforming everything for each element in that one iteration.

This PR also includes significant refactoring of the OASPathsTransformer tests to remove a lot of duplication.


mheusgit (approved):
Looks good

PR #167, opened 2022-05-04
Api 12891  docserver  create endpoint to make swagger friendly oas

mheusgit:
- Added ability to transform OAS into Swagger UI friendly versions that avoid performance pitfalls
- Added new renderConfig table to store each API versions associated render settings
- Updated OAS model to store new variant of OAS
- Added DataZip class to support OAS compression
- Added OAS key sorting function to mimic Postgres JSONB key behavior
- Updated a number of test cases


mheusgit:
PR cleanup is complete

dsteeve (dismissed):
nice work!

test/oas/transformer/oas.swaggerUI.transformer.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    },
+  );
+
+  describe('constructor', () => {
+    it('prepares for transformer successfully', () => {
+      function successfulTransformer() {
+        new SwaggerUITransformer(oneRingOAS, apiVersion);
+      }
+
+      void expect(successfulTransformer).not.toThrowError();
```
christineyost, line 59: I think void can be removed from all these `expect` statements

If you don't want to define a function for these you could also use an arrow function instead like
```
expect(() => new SwaggerUITransformer(oneRingOAS, apiVersion)).not.toThrowError();
```

But I also understand if you think the function definition is easier to read

Same in `dataZip.test.ts`

test/oas/transformer/oas.swaggerUI.transformer.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    it(`transforms OAS with reference errors`, () => {
+      function successfulTransformer() {
+        const sfTransformer = new SwaggerUITransformer(
+          invalidReferenceOAS,
+          apiVersion,
+        );
+        sfTransformer.transform();
+      }
+
+      void expect(successfulTransformer).not.toThrowError();
```
christineyost, line 173: This one could be
```
const sfTransformer = new SwaggerUITransformer(
  invalidReferenceOAS,
  apiVersion,
);

expect(() => sfTransformer.transform()).not.toThrowError();
```

test/oas/transformer/oas.swaggerUI.transformer.test.ts
```diff
@@ -0,0 +1,176 @@
...
+      const sfTransformer = new SwaggerUITransformer(
+        itemReferenceOAS,
+        apiVersion,
+      );
+      const transformedOAS = sfTransformer.transform();
+
+      expect(transformedOAS).toEqual(transformedItemReferenceOAS);
+    });
+
+    it(`transforms OAS with reference errors`, () => {
```
christineyost, line 164: What's the result of this transformation?

src/oas/transformer/oas.swaggerUI.transformer.ts
```diff
@@ -0,0 +1,420 @@
...
+  OASMediaType,
+  OperationKeys,
+} from '../oas.interface';
+import { cloneDeep } from 'lodash';
+import APIVersionModel from '../../api/version/apiVersion.model';
+import logger from '../../utils/logger';
+import ServerError, { DataResolveError, ErrorDetail } from '../../utils/errors';
+
+//SwaggerUITransformer prepares OAS to be consumed & rendered by Swagger-UI client by avoiding scenarios where gigantic schema trees get referenced & then rendered.
+//  Schemas with depth exceeding API versions configured pruneDepth are removed.  Unselected mediaTypes are also removed
```
christineyost, line 18: Are schemas pruned if their depth is equal to or exceeding the pruneDepth?

christineyost (commented):
Looks good to me!

Just have a suggestion on some of the test code.
Also had a question not to be addressed with this PR, but wondering if we should create a separate ticket

src/oas/oas.service.ts
```diff
@@ -274,6 +276,24 @@ class OASService {
...
       );
+
+      let sfTransformedSpec: OAS | undefined;
+
+      if (apiVersion?.renderConfig) {
+        const sfTransformer = new SwaggerUITransformer(
+          transformedSpec,
+          apiVersion,
+        );
+        sfTransformedSpec = sfTransformer.transform();
```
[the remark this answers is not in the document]
mheusgit: The other transformer is modeled as a service where this one is not.  To do that I think I would have to pass some kind of context object to most of the functions which I didn't want to do.

test/oas/transformer/oas.swaggerUI.transformer.test.ts
mheusgit: OAS still gets transformation performed but it contains reference errors.  Logic is expected to continue on even if one of the schemas is invalid.  Such problems require provider to correct it.

src/oas/transformer/oas.swaggerUI.transformer.ts
mheusgit: Should be exceeding.

christineyost: I think I'm a bit confused, or might be thinking about this wrong, but in the test when `pruneDepth = 3` the last schema level the OAS shows is level 2, so I thought everything at level 3+ is being pruned? And when `pruneDepth = 4` the last level is 3?

src/oas/oas.service.ts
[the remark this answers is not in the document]
christineyost: Oh yeah... the other transformer uses the GatewayService a couple times. We wouldn't want to spin up a new one of those every time

test/oas/transformer/oas.swaggerUI.transformer.test.ts
christineyost: Definitely agree. Would it be worth verifying that the transformed OAS is as expected in this test?

test/portalStore/portalStore.controller.test.ts
```diff
@@ -14,6 +14,14 @@ describe('PortalStoreController', () => {
...
+    sfTransformedSpec: 'Two Ring v2',
+    iteration: 1,
+  } as unknown) as OASModel;
+
+  const oas1 = ({
+    spec: 'One Ring v1',
+    transformedSpec: 'Two Ring v2',
+    sfTransformedSpec: 'Two Ring v2',
     iteration: 1,
   } as unknown) as OASModel;
```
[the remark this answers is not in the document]
mheusgit: They used to have a difference before PR.  Not needed anymore

src/oas/transformer/oas.swaggerUI.transformer.ts
mheusgit: Top most schema acts as a level.  The example component schemas are labeled for levels but not the starting location.
schema: {	<-- Level 1
  $ref: '#/components/schemas/topSchema',
},

topSchema: {	<-- Level 2
	description: 'Highest Schema',
	properties: {
	  deepProperty: {
		description: 'Should have pruned children',
		$ref: '#/components/schemas/deepSchema1',
	  },
	  shallowProperty: {
		description: 'Should be left alone',
		$ref: '#/components/schemas/shallowSchema',
	  },
	},
},

deepSchema1: {	<-- Level 3
	description: 'Level 2',
	properties: {
	  firstProperty: {
		$ref: '#/components/schemas/deepSchema2',
	  },
	},
},

shallowSchema: {	<-- Level 3
	description: 'Last Level',
	type: 'string',
	enum: ['additional', 'empty'],
},

mheusgit: If anything the descriptions could be updated

mheusgit: Description have been updated to be more clear.

christineyost: Got it! Thanks for updating those - I think that's really helpful!

PR #168, opened 2022-05-04
Api 15228 - Loast tables

ediiotero:
Story: [API-15228](https://vajira.max.gov/browse/API-15228)

Adding loast_result, loast_result_endpoint, and loast_result_exgroup models and interface
Updated Readme and added tests for the models


migrations/Migration20220503192847.ts
```diff
@@ -0,0 +1,19 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20220503192847 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('create table "loast_result" ("id" serial primary key, "created_at" timestamptz(0) not null, "updated_at" timestamptz(0) null, "server" varchar(255) null, "security_type" text check ("security_type" in (\'\')) not null, "oas_path" varchar(255) not null, "passes" int4 not null, "warnings" int4 not null, "failures" int4 not null, "count" int4 not null, "run_datetime" timestamptz(0) not null, "api_version_id" int4 not null);');
```
mheusgit, line 6: loast_result.security_type enum values are not correct.  I ran into this problem as well.

src/loast/loastResult/loastResult.model.ts
```diff
@@ -0,0 +1,76 @@
...
+import APIVersionModel from '../../api/version/apiVersion.model';
+import SecurityTypeEnum from '../../api/security/securityType.enum';
+import { LoastResult } from './loastResult.interface';
+
+@Entity({ tableName: 'loast_result' })
+class LoastResultModel extends BaseEntity {
+  @Property({ nullable: true })
+  server?: string;
+
+  @Enum({ type: 'SecurityTypeEnum' })
```
mheusgit, line 14: I think the fix for the invalid SQL is to change this to something similar to @Enum(() => SecurityTypeEnum}

src/loast/loastResultEndpoint/loastResultEndpoint.model.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { Optional } from '../../types';
+import { BaseEntity } from '../../utils/base.entity';
+import LoastResultModel from '../loastResult/loastResult.model';
+import MethodEnum from '../../method/method.enum';
+import { LoastResultEndpoint } from './loastResultEndpoint.interface';
+
+@Entity({ tableName: 'loast_result_endpoint' })
+class LoastResultEndpointModel extends BaseEntity {
+  @Enum({ type: 'MethodEnum' })
```
mheusgit, line 11: Same gotcha

migrations/Migration20220503192847.ts
mheusgit: Same story for loast_result_endpoint.method

dsteeve: what are the correct values here?  the confluence design doc might need updated. 
I would suggest we stick with the OAS swagger doc definition.

The type of the security scheme. Valid values are "apiKey", "http", "oauth2", "openIdConnect".

so technically we would just have an Enum of "apiKey" and "oauth2" for now?
but in the tests for this PR, it is using SecurityTypeEnum, which is defined with "bearer" and "apikey" which seems to be combining two OAS notions of SecurityType and SecurityScheme - we might need to pow-wow this thru.

christineyost: Oh dear, not very sure what we should do with this. I was thinking just "oauth" or "apikey" here. From a Lighthouse point of view, today we only support oauth or apikey security. If a provider uses oauth security, they might define it as an "oauth2" security scheme or as an "http" security scheme. Either way we need a token. The other problem is that some APIs define both "http" and "oauth2" security in the same OAS. I'm not sure what we would do with those if we're only allowing one value here

mheusgit (approved):
Looks good

src/loast/loastResultExgroup/loastResultExgroup.interface.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export interface LoastOutput {
+  // placeholder till loast output is created
+  results: Array<any>;
+}
+
+export interface LoastResultExgroup {
+  name: string;
+  errors: LoastOutput;
+  warnings: LoastOutput;
```
christineyost, line 11: I think these will end up being an array of objects like
```
{
  "message": "error/warning message here",
  "count": 3
}
```

Though we may end up adding a type field to this, so I'm ok with leaving a place holder for now. We should be sure to create a ticket to finish this up once the LOAST JSON output ticket is complete

christineyost (commented):
Looks good to me! I think once "PATCH" is added to the method enum this should be good to go. We should also create a follow up ticket for the loastResultExgroup errors and warnings

src/method/method.enum.ts
```diff
@@ -0,0 +1,8 @@
+enum MethodEnum {
```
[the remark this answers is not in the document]
ediiotero: thanks! Added PATCH

mheusgit:
Looks good

mheusgit (approved):

migrations/Migration20220503192847.ts
dsteeve: leaving security_type as just a string

migrations/Migration20220505193149.ts
```diff
@@ -0,0 +1,15 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20220505193149 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('alter table "loast_result" drop constraint if exists "loast_result_security_type_check";');
+    this.addSql('alter table "loast_result" alter column "security_type" type text using ("security_type"::text);');
+    this.addSql('alter table "loast_result" add constraint "loast_result_security_type_check" check ("security_type" in (\'bearer\', \'apikey\'));');
```
dsteeve, line 8: will this get automatically fixed since we aren't using an enum anymore?

src/loast/loastResult/loastResult.interface.ts
```diff
@@ -0,0 +1,14 @@
+import SecurityTypeEnum from '../../api/security/securityType.enum';
```
dsteeve, line 1: I think this import can be deleted now

README.md
```diff
@@ -550,6 +551,51 @@ A review of the OAS is done automatically during the <code> /refresh </code>
...
 </br></br>
 
+**Loast_Result**
+| Column Name  | Data Type & Length | Required | Format/Constraints | Example Value | Description |
+| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
+| id  | Integer  | Yes | Primary key | - | Unique ID of the loast_result |
+| created_at  | Timestamp with time zone  | Yes | YYYY-MM-DD HH:MI:SSZ | 2021-09-03 18:12:18+00 | Time loast_result was created |
+| updated_at  | Timestamp with time zone  | No | YYYY-MM-DD HH:MI:SSZ | 2021-09-03 18:12:18+00 | Last time loast_result was updated |
+| server  | Varchar (255)  | No | - | `https://sandbox-api.va.gov/services/vba_documents/{version}` | The server value that was used to run this LOAST test |
+| security_type  | Text  | Yes | "apiKey" or "oauth" | apiKey | Flag indicating whether apikey or oauth security was used for this LOAST test |
```
dsteeve, line 561: I"ve changed my mind since our earlier meeting today @ediiotero - 
After seeing we already have a securityType.enum in our code and after further discussing with @christineyost we decided a better name for this field would be "authentication_type"  This can then be an enum of "apikey" and "oauth". This follows along the lines of what this person wrote for different authentication types.  https://idratherbewriting.com/learnapidoc/docapis_more_about_authorization.html


This feels better to not confuse people.  Then all your tests you have will need to reference this new AuthenticationType.enum and not the securityTypeEnum that they are currently importing.

src/method/method.enum.ts
```diff
@@ -0,0 +1,9 @@
...
```
dsteeve, line 1: I think it would be better to have this live under the loast folder for now.

dsteeve (changes requested):
I think to make these changes, you will need to delete the current migration files and re-run the mikro-orm migration:up and then mirko-orm mirgration:create -- then the migration files should look correct in your PR.

test/loast/loastResult/loastResult.model.test.ts
```diff
@@ -0,0 +1,82 @@
+import APIModel from '../../../src/api/api.model';
+import APIVersionModel from '../../../src/api/version/apiVersion.model';
+import APIVersionStatusEnum from '../../../src/api/version/apiVersion.status.enum';
+import LoastResultModel from '../../../src/loast/loastResult/loastResult.model';
+import { LoastResult } from '../../../src/loast/loastResult/loastResult.interface';
+import SecurityTypeEnum from '../../../src/api/security/securityType.enum';
```
dsteeve, line 6: these would then be changed to be authenticationType.enum to not confuse people because I think this test will still work as is.

dsteeve (approved):
looks good to me!  Were the text fields going to be varchar still, or a separate ticket?

christineyost (approved):
Looks good to me!
As mentioned in the code comments, we can update the LoastOutput interface when API-15225 is worked

ediiotero:
I added updating the LoastOutput interface to the AC for API-15225

PR #169, opened 2022-05-04
Api 12919  s3 needs a place to store swagger friendly oas

mheusgit:
- Updated OAS logic to include the SwaggerUI friendly variant OAS during OAS uploads

Has a dependency on https://github.com/department-of-veterans-affairs/lighthouse-docserver/tree/API-12891--docserver--Create-endpoint-to-make-swagger-friendly-OAS


src/oas/oas.service.ts
```diff
@@ -274,6 +276,24 @@ class OASService {
...
+
+      if (apiVersion?.renderConfig) {
+        const sfTransformer = new SwaggerUITransformer(
+          transformedSpec,
+          apiVersion,
+        );
+        sfTransformedSpec = sfTransformer.transform();
+      }
+
+      //Temp fix: sfTransformedSpec is stored in Bytea format in postgres
```
dsteeve, line 290: I see why this is here for consistency of current behavior, but like we discussed this morning, a better option is to not have this behavior that jsonb and postgres does. future work for sure.

migrations/Migration20220426174903.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  async up(): Promise<void> {
+    this.addSql('create table "render_config" ("id" serial primary key, "created_at" timestamptz(0) not null, "updated_at" timestamptz(0) null, "media_types" varchar(255) not null, "prune_depth" int4 not null);');
+
+    this.addSql('alter table "api_version" add column "render_config_id" int4 null;');
+    this.addSql('alter table "api_version" add constraint "api_version_render_config_id_unique" unique ("render_config_id");');
+
+    this.addSql('alter table "api_version" add constraint "api_version_render_config_id_foreign" foreign key ("render_config_id") references "render_config" ("id") on update cascade on delete set null;');
+
+    this.addSql('alter table "oas" add column "sf_transformed_spec" bytea not null;');
```
dsteeve, line 13: should this be nullable, since the existing records won't have this data?

src/oas/oas.model.ts
```diff
@@ -37,16 +44,32 @@ class OASModel extends BaseEntity {
   })
   apiVersion!: APIVersionModel;
 
+  sfTransformedSpec!: OAS; //Shadow property for zippedSFTransformedSpec
+
+  @OnInit()
```
dsteeve, line 49: it looks from the mikro orm documentation that OnInit might get called twice in some situations.  I don't think it would be a huge deal if that happened, but something to be aware of.

dsteeve (dismissed):
great work! Just one question on the "not null" column being added.  That was the only potential issue I could see when deploying to a non-empty database.

src/oas/oas.service.ts
mheusgit: A ticket (API-15748) to make this fully configurable has been created.

migrations/Migration20220426174903.ts
mheusgit: This was switched to nullable in the sibling ticket.  It has to be nullable to work.

src/oas/oas.model.ts
mheusgit: Funny you mention it. @OnLoad is the call we want long term but that annotation appears to require a later Mikro-ORM version

dsteeve: ah, yes I saw that too.

migrations/Migration20220426174903.ts
dsteeve: ok, cool.

src/oas/oas.service.ts
dsteeve: ok, thank you!

mheusgit:
Merge hold until "https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/167" occurs first

christineyost (commented):
Looks good to me!

I think we'll also need to update Metadata json blobs to point at openapi-sf.json instead of openapi.json so that the swagger friendly OASs show up on the portal. Do we have another ticket for that?

mheusgit:
Dependency ticket has been merged.  This ticket was made allot smaller once the conflicting changes were addressed

christineyost (approved):
Looks good to me!

PR #170, opened 2022-05-05
Api 15210  spike  doc server data recreate/retrieval

mheusgit:
Added profile against  pgAdmin service to keep consistency with deployed environments


dsteeve (approved):
looks like a good thing to try

PR #171, opened 2022-05-27
API-15226/  create loast config model and database table   paramstore poc

mheusgit:
- Added new service support for calling AWS parameter store
- Included a temporary section for driving the POC under the serviceInjector.ts


test/util/paramstore.service.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      region: cfg.AWS_REGION,
+      credentials: {
+        accessKeyId: cfg.AWS_S3_ACCESS_KEY,
+        secretAccessKey: cfg.AWS_S3_SECRET_KEY,
+      },
+    });
+  });
+
+  it('builds a localstack with SSM parameter prepopulated if the environment is local', () => {
+    cfg.SSM_PARAMETERS = `{"/paramPath/param1": "fakeValue1","/paramPath/param2": "fakeValue2"}`;
```
christineyost, line 48: If this environment has two params will the clientRequestMock be called twice? I'd suggest removing the second parameter here or adding an expect for the second

christineyost (approved):
Looks good to me! Added a  small suggestion on a test, but I'm ok with it as is since it's testing code for local development only

test/util/paramstore.service.test.ts
mheusgit: Yeah.  This can be cleaned up.  Will deal with it in the main ticket (none POC)

PR #172, opened 2022-06-02
API-15226  create loast config model and database table

mheusgit:
- loast_config table is added to the Docserver database
- Changes include new loastConfig model and associated changes to api_version model
- Documentation updated
- Switched the column on the data model for better readability


docs/oas-database-structure.mmd
```diff
@@ -1,87 +1,138 @@
 erDiagram
 
 api {
-  integer id
-  timestamp created_at
-  varchar name
-  timestamp updated_at
-  varchar gateway_id
-  integer metadata_counter_id
+  id integer
```
dsteeve, line 4: really like this reformatting!  makes the diagrams more readable.  Thank you!

src/api/loastConfig/loastConfig.interface.ts
```diff
@@ -0,0 +1,13 @@
+import AuthenticationTypeEnum from '../../loast/loastTypes/authenticationType.enum';
+
+export interface LoastConfig {
+  server?: string;
+  securityType: AuthenticationTypeEnum;
```
dsteeve, line 5: bummer.. in this PR https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/168 we had decided to call this field "authenticationType" instead of securityType since that name was already used and had different values.  We forgot to change this ticket accordingly, and I also noticed in that PR https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/168 the readme wasn't updated for the "authenticationType" for the loast_result table.  That should get fixed too. 

Basically "authenticationType" is either apikey or oauth.  ("how will the consumer be authenticating")
And "securityType" is the value in the header.  Either an apikey or bearer token. (what is the type of token)

migrations/Migration20220524234045.ts
```diff
@@ -0,0 +1,15 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20220524234045 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('create table "loast_config" ("id" serial primary key, "created_at" timestamptz(0) not null, "updated_at" timestamptz(0) null, "server" varchar(255) null, "security_type" text check ("security_type" in (\'oauth\', \'apikey\')) not null, "oas_path" varchar(255) not null);');
```
dsteeve, line 6: I think this should be authentication_type instead of security_type and the field should be varchar instead of text. 

In looking at the loast_result table, that is authentication_type, but it is a text field and it should be like varchar(16).  

Either we can create another ticket or fix with this one?  
loast_result_table:  authenticationType should be varchar(16)
loast_result_endpoint: method should be varchar(16)

dsteeve (commented):
It was all done correctly based on the original ticket, but I think we need to get this securityType, authenticationType choice consistent.

docs/oas-database-structure.mmd
mheusgit: Oh yeah.  I didn't mention the diagrams had the column names vs types switched

migrations/Migration20220524234045.ts
mheusgit: for data model changes... gonna go with fixing this ticket

dsteeve: sound good to me!

mheusgit: Not following on why we would make this tables column a varchar but have loast_result.authentication_type use text.  They should bot match.  

End of day the value from loast_config gets copied over to loast_result so we have a record of what was run.

mheusgit:
Docs and all have been switched from "securityType" to "authenticationType"

dsteeve (dismissed):
looks great! 

I think I updated all the places on the confluence page that the ticket referenced, but you might want to double check my work there. (I know that diagram is still outdated on confluence)

migrations/Migration20220524234045.ts
dsteeve: right, I vote for varchar on the loast_result.authentication_type.  Maybe my comment isn't clear.

mheusgit: If you want the text converted to varchar we should also change the following columns
- loast_result_endpoint.method
- api_version.security_type
- api_version.status
- oas.status

mheusgit: What is the driving reason for switching column types for constrained data?

dsteeve: after reading some more, I see that this is a mikro-orm deficiency in how it handles enums with postgres
https://github.com/mikro-orm/mikro-orm/issues/2764

So, instead of trying to work around our tools, and having inconsistency with the datatypes already set as text, let us continue down the path we are on and stay at least be consistent.  I don't think we should have enough data to matter, and hopefully postgres will be smart about things since the text value length is small. 

Ideally, the "status", "security_type", "authentication_type", "method" should have all been reference tables (or maybe enum types in postgres).  Unfortunately, because of how this ORM operates, and that we haven't been creating reference tables and have been using this handy "Enum" type, the ORM created text fields.

When a database manages a text field, that text field can be of any size, therefore the storage of it is less performant than that of a varchar where the max size is known in a varchar.  In addition, if you were ever to put an index on a text field, or a large varchar field, it is far less efficient.  
Since we already know what we want the values to be, it makes much more sense to make it a varchar of limited size for these fields since for some of these fields we may wish to add an index at some point. 
 
 One can argue that for our tiny amount of data, that it is not going to matter.  Here is another article that gives me some hope that the db will perform just fine for our needs here since our text values are tiny in size. https://hakibenita.com/sql-medium-text-performance

dsteeve (approved):
Looks good!

PR #173, opened 2022-06-02
Api 13412  doc server request retry attempts

mheusgit:
- Added Axios-retry module to improve external request behavior
- Added maxAttempts config against S3 client
- Added two new configs to control retry attempts and additional logging connected to retry failures


ediiotero (approved):
Looks good to me

src/api/version/apiVersion.service.ts
```diff
@@ -22,14 +31,27 @@ export const responseInterceptor = (response: AxiosResponse): AxiosResponse => {
   return response;
 };
 
+export const errorInterceptor = (error: AxiosError): Promise<AxiosError> => {
+  logger.info(error.message);
```
dsteeve, line 35: should this be logger.error or does this then get caught higher up and that writes an error message to the log? Just want to make sure if we did a search on errors in the log we pick these up.

dsteeve (approved):
cool! Can't wait to see in action.

PR #175, opened 2022-06-07
Set updatedAt to match createdAt on entity creation

christineyost:
This PR is for [API-15021](https://vajira.max.gov/browse/API-15021). The updatedAt field will now be populated with the current time on entity creation. updatedAt is no longer nullable. This change includes updating existing data in the database migration - if updatedAt is currently null it will be set to the same value as createdAt.


migrations/Migration20220603214843.ts
```diff
@@ -0,0 +1,62 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20220603214843 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('update "render_config" set "updated_at" = "created_at" where "updated_at" is null;');
```
mheusgit, line 6: That works

migrations/Migration20220603214843.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.addSql('alter table "render_config" drop constraint if exists "render_config_updated_at_check";');
+    this.addSql('alter table "render_config" alter column "updated_at" type timestamptz(0) using ("updated_at"::timestamptz(0));');
+    this.addSql('alter table "render_config" alter column "updated_at" set not null;');
+
+    this.addSql('update "gateway_config" set "updated_at" = "created_at" where "updated_at" is null;');
+    this.addSql('alter table "gateway_config" drop constraint if exists "gateway_config_updated_at_check";');
```
mheusgit, line 12: These constraints that were dropped.  I don't see them in the migration files at any point in time

mheusgit (approved):
Looks good

mheusgit:
Delay this tickets merge until after 16:00 EST since we have a MR

dsteeve (approved):
looks good!

PR #181, opened 2022-06-16
Regenerate and upload all metadata

ediiotero:
Story: [API-16715](https://vajira.max.gov/browse/API-16715)

The way this currently works is it will run generateMetadataAndUpload() on every api on the list and stop if an error is thrown showing the thrown error as the response.


src/portalStore/portalStore.controller.ts
```diff
@@ -13,13 +15,23 @@ class PortalStoreController {
...
+  constructor(metadataService: MetadataService, oasService: OASService, apiService: APIService) {
     this.metadataService = metadataService;
     this.oasService = oasService;
+    this.apiService = apiService;
+    this.initializeRoutes();
     this.initializeSecureRoutes();
   }
 
+  public initializeRoutes(): void {
+    this.router.get(
```
mheusgit, line 29: SecureRoute & use POST

src/portalStore/portalStore.controller.ts
```diff
@@ -13,13 +15,23 @@ class PortalStoreController {
...
+      '/metadata-all',
```
mheusgit, line 30: Was hoping the types of data being refreshed would be in JSON body.
{ dataType: "metadata,oas,etc" }  
Approach also would allow removal of outdated folders if dataType was not provided

test/portalStore/portalStore.controller.integration.ts
```diff
@@ -66,6 +66,26 @@ describe('PortalStoreController', () => {
       .auth(cfg.ADMIN_USERNAME, cfg.ADMIN_PASSWORD);
   });
 
+  describe('unsecured routes', () => {
```
mheusgit, line 69: secure

test/portalStore/portalStore.controller.integration.ts
```diff
@@ -144,5 +144,24 @@ describe('PortalStoreController', () => {
...
+        process.env = {
+          ...PREV_ENV,
+          OAS_ENV: EnvType.PROD,
+          NODE_ENV: NodeEnvType.TEST,
+        };
+        configRefresh();
+      });
+
+      it('regenerates and uploads all metadata with 200', async () => {
+        const response = await request.get(
```
mheusgit, line 160: HTTP method should be a POST

mheusgit (approved):
Looks good

PR #185, opened 2022-06-27
API-14299  provider dashboard  api endpoint status (loast detailed report header and passes) 2

mheusgit:
- Added new Loast controller endpoint that returns sample test results


peyton-lucas (approved):
LGTM!

PR #186, opened 2022-06-27
Add an endpoint for storing LOAST results

christineyost:
This PR is for [API-15229](https://vajira.max.gov/browse/API-15229). It adds an endpoint (`POST /docserver/apis/<name>/versions/<version>/loast/results`) that stores LOAST results in the Docserver loast_result, loast_result_endpoint, and loast_result_exgroup tables.

Added a sample LOAST result JSON to `dev/loastOutput/loast.json`

Made the authenticationType field in the loast_result table nullable in case the authentication type in the result does not match our expected enum. We will store the result with a null authentication type and log a warning.

Made the method field in the loast_result_endpoint table nullable in case an HTTP method in the result does not match our expected enum. We will store the result with a null method and log a warning.

Also wanted to note that this is a first attempt at handling transactions in Docserver. [API-17040](https://vajira.max.gov/browse/API-17040) is currently being worked on and once that pattern is in place this will be refactored to follow the new pattern.


dsteeve (dismissed):
Christine and I went thru an in-person code review on this today since she is on PTO next week.  

We discussed some ideas to break out the code better in this file
https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/186/files#diff-a483737d980ea3dd11e217152a8b6f3b45780df4f6895a90e91d54645636dc52


That work can go into a different ticket once the ideas from API-17040  is brought in.  

So approving this knowing we can make a few improvements later.

stevelong00:
Looks good to me

stevelong00 (approved):
LGTM!

src/loast/loast.controller.ts
```diff
@@ -66,12 +88,24 @@ class LoastController {
...
+      );
+      response.sendStatus(201);
+    } catch (err) {
+      next(err);
+    }
+  };
+
   getResultById: RequestHandler = (request, response, next) => {
     try {
       // TODO Call the loastService to get the data from the db once that is in place
```
mheusgit, line 106: TODO can be removed now since this is implemented

src/loast/loast.service.transactional.ts
```diff
@@ -0,0 +1,192 @@
...
+      oasPath,
+      totalPass,
+      totalWarn,
+      totalFail,
+      totalRun,
+      runDateTime: new Date(runDateTime),
+      apiVersion,
+    };
+
+    return new LoastResultModel(
```
mheusgit, line 124: When would we not have the server.  Maybe a comment is needed

src/loast/loast.service.transactional.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  constructor(orm: MikroORM) {
+    this.entityManager = orm.em;
+  }
+
+  public async persistLoastResult(
+    apiVersion: APIVersionModel,
+    loastRequestBody: LoastRequestBody,
+  ): Promise<LoastResultModel> {
+    const loastResultModel = this.createLoastResultModel(
```
mheusgit, line 33: A verbose log just before this line to output the loastRequestBody is needed

src/loast/loastResult/loastResult.interface.ts
```diff
@@ -3,13 +3,13 @@ import APIVersionModel from '../../api/version/apiVersion.model';
 
 export interface LoastResult {
```
mheusgit, line 4: I believe these set of interfaces could be combined with the loast.interfaces.ts code

mheusgit (dismissed):
The verbose log is needed at a minimum

src/loast/loast.interface.ts
```diff
@@ -0,0 +1,63 @@
...
+  runDateTime: string;
+}
+
+export interface ExampleGroupResult {
+  errors: LoastError[];
+  warnings: LoastWarning[];
+}
+
+export interface LoastError {
+  message: string;
```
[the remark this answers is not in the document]
christineyost: The only optional field I see is `server`. That field is not required by LOAST - it is optional if the OAS only has one server. I'll add that comment as mentioned below

Having the LoastError/LoastWarning interface separated mirrors the LOAST output interface: https://github.com/department-of-veterans-affairs/lighthouse-oas-tests/blob/master/src/utilities/structured-output/structured-output.interface.ts. I was intending to create these interfaces for the LOAST output specifically, not Docserver output

src/loast/loast.service.transactional.ts
christineyost: Added a comment

src/loast/loast.service.transactional.ts
christineyost: The request body is already logged when the request comes in. Do we need an additional verbose log?

src/oas/oas.interface.ts
```diff
@@ -316,4 +316,17 @@ export const OperationKeys = [
   'trace',
 ];
 
+export type OASSecuritySchemeType =
+  | 'apiKey'
+  | 'http'
+  | 'oauth2'
+  | 'openIdConnect';
+
+export enum OASSecuritySchemeTypeEnum {
```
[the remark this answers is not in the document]
christineyost: Removed the type

src/loast/interfaces/loast.input.interface.ts
```diff
@@ -0,0 +1,63 @@
+export interface LoastRequestBody {
+  id: string;
+  config: TestConfig;
+  results: APIResults;
+}
```
dsteeve, line 5: This is a subjective naming suggestion.  Not sure I should even comment,but for what its worth, I feel like there is a mix of "test" and "api" going on here and it feels confusing.  

Not sure if this is better, but what if this was :
`config: LoastRunConfig;
 results: LoastRunResults
 `

src/loast/interfaces/loast.input.interface.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export function isAPISummary(
+  input: APISummary | OperationResult,
+): input is APISummary {
+  return (input as APISummary).runDateTime !== undefined;
+}
+
+type APIResults = {
+  apiSummary: APISummary;
```
dsteeve, line 20: going along with the naming above suggested, what if this was 
LoastSummary and LoastResults instead of APISummary and APIResults


not sure if it is more readable or not.  What I'm worried about is that we have an API table in docserver, and want to make sure that this is obvious as being part of testing the API and not some summary description of an API.

dsteeve (commented):
looks good - like what you have done, but just hung up on some naming that we can quickly talk thru.

dsteeve (approved):
Looks great to me!

stevelong00 (commented):
Looks good!

PR #187, opened 2022-07-20
Api 17040/db tranmgmt

dsteeve:
This is a draft idea of how to do propagation for docs moving from dev to staging or staging to prod.

It introduces the idea of a transaction manager to do explicit transactions.  It also introduces the idea of a manager that a controller calls, this avoids circular reference issues with services knowing about other services. (as an example:  apiService would need to know about propagationService, and vice-versa, that is a circular reference that introducing apiManager fixed)

Goal of this ticket was to write some code that does explicit database transaction management and use it in a propagation proof of concept.


src/api/propagation/propagation.manager.ts
```diff
@@ -0,0 +1,129 @@
...
+      ]);
+    }
+
+    const apiVersion: number = requestBody.majorVersion;
+    if (apiDoesExist && !!apiModel) {
+      const apiVersionDoesExist: boolean = await this.apiVersionService.doesAPIVersionExist(
+        apiModel,
+        apiVersion,
+      );
+      if (apiVersionDoesExist) {
```
mheusgit, line 69: I wouldn't think to be an error.  We would just update the API versions configurations that are different

src/api/api.manager.ts
```diff
@@ -0,0 +1,53 @@
...
+   * create a new API
+   * This could be done via propagation or by just creating a row in the api table
+   * @param api
+   */
+  public async createAPI(requestBody: APIRequestBody): Promise<APIModel> {
+    const doesExist: boolean = await this.apiService.doesAPIExist(
+      requestBody.name,
+    );
+    let apiModel: APIModel;
+    if (doesExist) {
```
mheusgit, line 37: To capture previous conversation: Seems the two managers needed to have awareness of one another.  Maybe this API manager logic should ultimately be in PropagationManager instead

src/api/api.controller.ts
```diff
@@ -79,22 +82,17 @@ class APIController {
...
-    const reqBody = request.body as API;
-    const doesExist = await this.apiService.doesAPIExist(reqBody.name);
-    if (doesExist) {
-      response.sendStatus(422);
-      return;
-    }
+    const reqBody = request.body as APIRequestBody;
     try {
-      await this.apiService.createAPI(reqBody);
+      await this.apiManager.createAPI(reqBody);
```
mheusgit, line 87: I didn't catch this earlier.  Wouldn't the propagation logic fall under a new endpoint?

mheusgit:
@dsteeve We should be able to close this PR.  Approach for propagation went a different direction so I don't believe this will be needed.

PR #188, opened 2022-07-22
API-16219/  doc server  create an endpoint for retrieving loast results list

mheusgit:
- Change implemented loast result list endpoint.
- Removed hardcoded sample result summary page
- Included support for pagination of result sets


src/loast/loastResult/loastResult.filter.ts
```diff
@@ -0,0 +1,15 @@
+import { LoastResultResponse } from '../interfaces/loast.output.interface';
+import LoastResultModel from './loastResult.model';
+
+const filterLoastResultResponse = (
```
dsteeve, line 4: curious why this is called "filter".  What is being filtered with it?  
Is it more of a tally?

ediiotero (approved):
lgtm

dsteeve (approved):
I think this looks great Marc!
I think loastResult.tally vs loastResult.filter might be a better name.  thoughts?  
I generally think of filter as something that is filtering things out -- like "I only want errors".

src/loast/loastResult/loastResult.filter.ts
mheusgit: The naming comes from the other "filters" that message the data.  api.filter.ts, gatewayConfig.filter.ts, oas.filter.ts are some examples.  We can rename all of them if needed.

PR #191, opened 2022-07-27
Update nodemon package to latest version

christineyost:
This PR is for [API-17361](https://vajira.max.gov/browse/API-17361). It updates the `nodemon` package to the latest version in order to remove the problematic `got` dependency.

Confirmed Docserver still builds, tests still pass, and the Docserver container still restarts when source code changes.


mheusgit (approved):
Fewer places to fail now

dsteeve (approved):
looks good!

PR #195, opened 2022-08-01
API-17356/ Change Kong instance to be db-less for local dev setup

mheusgit:
- Switched local dev environment Kong to DB-less mode and added corresponding kong.yaml to load setup instead
- Removed four services connected to local dev Kong DB migrations & initial DB setup as they were no longer needed
- Updated readme


dsteeve (approved):
we did an in-person code review yesterday.  It all looks good.

PR #196, opened 2022-08-03
API-15744  ../oas endpoint shouldn't retrieve all the OAS blobs

ediiotero:
Story: [API-15744](https://vajira.max.gov/browse/API-15744)

Added `lazy` property to `spec` and `transformedSpec` in the OASModel.

Now when using the  `GET: /docserver/apis/<apiName>/versions/<apiVersion>/oas` endpoint the OAS is removed to speed up the response.

Both `GET /docserver/apis/<apiName>/versions/<apiVersion>/oas/<oasIteration>` and
`GET /docserver/apis/<apiName>/versions/<apiVersion>/oas/latest` still work as expected


src/oas/oas.model.ts
```diff
@@ -21,10 +21,10 @@ class OASModel extends BaseEntity {
...
 
-  @Property({ type: JsonType })
+  @Property({ type: JsonType, lazy: true })
   spec!: OAS;
 
-  @Property({ type: JsonType })
+  @Property({ type: JsonType, lazy: true })
   transformedSpec!: OAS;
 
   @Property({
```
christineyost, line 30: Should this field also be lazy?

christineyost (commented):
Looks good to me! Just wondering if the sf_transformed_spec should also be lazy?

src/oas/oas.model.ts
ediiotero: Yes, that makes sense.  Made the change.

mheusgit (approved):
Looking forward to testing this

PR #198, opened 2022-08-11
Fix for getting OAS by version + iteration and getting latest

ediiotero:
GET /docserver/apis/<apiName>/versions/<apiVersion>/oas/<oasIteration>
GET /docserver/apis/<apiName>/versions/<apiVersion>/oas/latest

Should now respond with the OAS


mheusgit:
Pretty sure this needs a corresponding change to the oas.controller.ts

mheusgit:
Looks good

mheusgit (approved):
Everything appears to be in place

PR #200, opened 2022-08-22
Api 18208/fix tsc errors in test code

stevelong00:
* removed unnecessary mockRequest interface
* configured eslint to ignore dist-test folder in addition to dist


mheusgit (approved):
That should help

PR #201, opened 2022-08-23
Create DatabaseService and update LoastTransactionalService to use it

christineyost:
This PR is for [API-17659](https://vajira.max.gov/browse/API-17659). It adds the DatabaseService and updates LoastTransactionalService to use it. Tests were also updated and added as needed.


src/loast/loast.service.transactional.ts
```diff
@@ -83,24 +86,25 @@ class LoastServiceTransactional {
       },
     );
 
-    const em = this.entityManager.fork();
-    await em.begin();
+    const loastResultsModels: (
```
dsteeve, line 89: it might be more clear to call this "loastModels" so it isn't confused with being just "LoastResultModel" -- but that is more of a style and opinion thing.

test/database/database.service.integration.ts
```diff
@@ -0,0 +1,43 @@
...
+  beforeEach(async () => {
+    await wipeDatabaseRows(orm.em);
+  });
+
+  afterAll(async () => {
+    await orm.close(true);
+  });
+
+  describe('executeTransaction', () => {
+    it('stores entities in the database', async () => {
```
dsteeve, line 25: would it make sense (is it feasible) to add a second test there that does a rollback? 
I think you can do it by having a second test it('rollsback on error') 
and in there have a name that exceeds the length of the database table field.  Like "somereallyreallyreallylongnameforanAPIthatNoOneWouldEverDo,ButWeAreDoing" and put that in the two_ring name and then it should rollback that transaction.  
curious if that would work.

dsteeve (approved):
this looks good as it is right now, just had a couple comments/ideas.

test/database/database.service.integration.ts
christineyost: Added a test for this

dsteeve (approved):
looks great!

PR #202, opened 2022-08-29
Api 16584  doc server   jest db migration timeout error

mheusgit:
- Created new initial migration
- Moved old migrations down to a subfolder
- Removed problematic jest timeout setting


mheusgit:
Merge to be performed after 9/5

dsteeve (approved):
Marc did an in-person code review this week with the team.  I also tried it locally with a db existing and with a whole new db.  Both ways seemed to work as expected.

mheusgit:
Manual db snapshots created prior to merging.

PR #204, opened 2022-09-09
API-18942/  add additional fields to api table

mheusgit:
- Added displayName, displayDescription, and signupId to the API table


dsteeve (approved):
looks good to me.

mheusgit:
PR needs a small change for varchar sizing

dsteeve (approved):
Marc and I discussed offline the need for name and signup_id to be shorter - since it is possible they could get an index on them some day.  
He found the ORM magic, and made that change. Thanks Marc!

PR #205, opened 2022-09-14
Api 14296/size limit gateway response

ediiotero:
Story: [API-14296](https://vajira.max.gov/browse/API-14296)

Update Responses in OAS transformer to support the size limit plugin

I used a 413 but open to feedback from the team


test/fixtures/gatewayResponses.ts
```diff
@@ -47,4 +47,20 @@ export const gatewayResponses = {
       },
     },
   },
+  '413': {
```
mheusgit, line 50: Not sure why we are not reusing the exports from "src/oas/transformer/resources/gatewayResponses.ts" for the fixture and friends

src/oas/transformer/resources/gatewayResponses.ts
```diff
@@ -50,3 +50,20 @@ export const response429: OASResponse = {
     },
   },
 };
+
+export const response413: OASResponse = {
+  description: 'Payload too large',
```
mheusgit, line 55: It so odd that shortly after this PR was created but not merged that DocServer started generating "413" type errors for LOAST results in [API-18995](https://vajira.max.gov/browse/API-18995).  

Just leaving this comment for tracking.

mheusgit (approved):

test/fixtures/gatewayResponses.ts
mheusgit: I'll let ticket creator decide if this should be cleaned up.

src/oas/transformer/resources/gatewayResponses.ts
dsteeve: that is very curious?

test/fixtures/gatewayResponses.ts
dsteeve: I would say that makes sense as well.  It is just exporting constants.

src/oas/transformer/resources/gatewayResponses.ts
mheusgit: No.  Its just to avoid confusion when the bug fix goes in for Express 413.  It not connected to this change.

dsteeve (changes requested):
if we have the time, we should do the test fix.  Also blocking this until the actual Kong changes are closer to production, otherwise we will potentially have this 413 added to all the OAS's in production docserver before Kong changes are actually live.

dsteeve (approved):
I think this can go in now, even if the gateway change lags a bit.

PR #206, opened 2022-09-14
API-18942/ Add additional fields to api table

mheusgit:
- Updated signupId to be optional


dsteeve (approved):
does the db schema already allow for null?

mheusgit:
alter table "api" add column "display_name" varchar(40) null, add column "display_description" varchar(255) null, add column "signup_id" varchar(40) null;

Yes

PR #208, opened 2022-09-22
API-15901/oas cleanup

ediiotero:
Story: [API-15901](https://vajira.max.gov/browse/API-15901)

Added a batch cleanup route for APIs with old OASs 
`docserver/oas/cleanup?oasStatus=inactive&olderThan=120`


mheusgit:
I'd approve this but I wrote some of it

src/oas/oas.service.ts
```diff
@@ -436,6 +436,42 @@ class OASService {
...
+
+    for (const apiVersion of apiVersions) {
+      const apiVersionStr = `${apiVersion.api.name} / v${apiVersion.majorVersion}`;
+
+      try {
+        logger.info(
+          `Batch OAS cleanup: Deleting old iterations: ${apiVersionStr} `,
+        );
+
+        totalDeleted += await this.deleteOAS(apiVersion, oasStatus, expireDate);
```
christineyost, line 454: Would it be helpful to log how many were deleted for each APIVersion were deleted, just after this delete call?

test/oas/oas.service.test.ts
```diff
@@ -540,6 +540,29 @@ describe('OASService', () => {
...
+        true,
+        APIVersionStatusEnum.CURRENT,
+        true,
+      );
+      apiVersionService.getAPIVersionsList = jest
+        .fn()
+        .mockResolvedValue([apiVersionModel, apiVersionModel2]);
+
+      service.deleteOAS = jest.fn().mockResolvedValue(2);
+      await service.batchOAScleanup(StatusEnum.INACTIVE, new Date());
```
christineyost, line 560: Might be nice to also assert what this call returns? I think it would be 4 in this case?

christineyost (commented):
Looks good.

Would it be possible to add a test to `oas.service.integration.ts` for the new `batchOASCleanup` method?

stevelong00 (approved):
Looks good to me!

PR #211, opened 2022-10-03
API-20301/ DocServer: SF OAS generation excluding transformed added mediaTypes

mheusgit:
- Added a fallback check for mediaTypes.  If no media type matches the setting in a render config then the very first result will still be allowed.


ediiotero (approved):
Looks good

mheusgit:
Merge delayed due to recent ECS changes

PR #213, opened 2022-10-11
API-20393/fix dependabot alerts

mheusgit:
- Updated LOAST to correct two insecure modules (parse-path & parse-url)
- Updated Jest due to LOAST update causing Spectral nimma failure.  Associated to NPM version change
- Updated passport due to insecure module issue


christineyost (approved):
Seems like a bit of the `runAPIVersionBatch` logic needs to be re-thought, but this should be fine for fixing dependabot alerts

mheusgit:
> Seems like a bit of the `runAPIVersionBatch` logic needs to be re-thought, but this should be fine for fixing dependabot alerts

Agreed about need of some rework.  If anyone is curious the direction rework is headed goes back to API-20279 and the first comment on ticket.

PR #214, opened 2022-10-13
API-15901/ Auto Cleanup Scheduler 

ediiotero:
Story: [API-15901](https://vajira.max.gov/browse/API-15901)

Updating the scheduler endpoint for the new DELETE cron job


src/scheduler/scheduler.schema.ts
```diff
@@ -3,5 +3,6 @@ import Joi from '@hapi/joi';
 export const schedulerSchema = Joi.object()
   .keys({
     task: Joi.string().required(),
+    olderThan: Joi.string(),
```
christineyost, line 6: This should probably match the `olderThan` schema for `oasIterationDeleteSchema`:

```
olderThan: Joi.number().min(0).max(180), //Permitted range 0-180
```

christineyost (commented):
Looks good to me! Just have a question about the Joi schema

src/scheduler/scheduler.controller.ts
```diff
@@ -25,11 +25,14 @@ class SchedulerController {
   executeTask: RequestHandler = (request, response, next) => {
     try {
       const { task } = (request.query as unknown) as ScheduledJobQuery;
-
+      const cleanupTask = '/oas/cleanup?oasStatus=inactive&olderThan=120';
```
mheusgit, line 28: cleanupTask isn't needed.

Line at 32 should be simplified and future proofed by using line below.
"request.method = task.includes('/oas/cleanup') === cleanupTask ? 'DELETE' : 'POST';"

This way if we need to change how schedule works it doesn't require this code to be changed again.

src/scheduler/scheduler.schema.ts
ediiotero: Without it I'm getting:
```
{
    "errors": [
        "\"olderThan\" is not allowed"
    ]
}
```

mheusgit: Yeah.  The scheduler schema looks like it will get to become a dumping ground for the combined cronjob schemas.  Not sure there is a better approach.

src/scheduler/scheduler.controller.ts
christineyost: Since we're doing it this way should we just call the scheduler by `GET .../docserver/scheduler?task=/oas/cleanup`, especially since olderThan=120 and oasStatus=inactive are being defaulted in no matter what the user passes in? I think that way we wouldn't need `olderThan` on the scheduler schema?

Or should we use the `olderThan` and `oasStatus` values passed in the `task` query parameter?

ediiotero: @christineyost that is a good point. @mheusgit thoughts?

src/scheduler/scheduler.controller.ts
```diff
@@ -29,7 +29,10 @@ class SchedulerController {
       //Prepare to forward to router to handle task
       request.url = `/docserver${task}`;
       request.originalUrl = request.url;
-      request.method = 'POST';
+      request.method = task.includes('/oas/cleanup', 0) ? 'DELETE' : 'POST';
+
+      if (request.method === 'DELETE')
+        request.query = { oasStatus: 'inactive', olderThan: '120' };
```
mheusgit, line 35: Team requested this be made dynamic

src/scheduler/scheduler.controller.ts
mheusgit: I'd prefer the two parameters originate from the cronjob so we only need to update the job's repo if we need to change the frequency or behavior.

mheusgit (dismissed):
Looks good to me

christineyost:
Eddie and I discussed trying to change some things. This current solution will only work if the query parameters inside the task query parameter remain in this order. To fix the problem, the `&` in the task query parameter needs to be URL encoded. If that is done `olderThan` will not need to be added to the Scheduler's request schema. With this change the way `oasStatus` and `olderThan` are fetched from the task query parameter will also need to be changed

dsteeve (approved):
looks good!

mheusgit:
Looks good

PR #216, opened 2022-10-17
API-18204 - Add endpoint and component statistics

christineyost:
This PR is for [API-18204](https://vajira.max.gov/browse/API-18204). It adds endpoint and component statistics to Slack messages.

Endpoint statistics are generated by comparing the new transformed OAS's operationIds to the old transformed OAS's operationIds. Since the `OASTransformer` generates operationIds based on path and operation method, these should always be present for all operations on a transformed OAS, and they will be consistent between iterations.

Component statistics are generated by comparing the new transformed OAS's `components.schemas` keys to the old transformed OAS's `components.schemas` keys.


src/differ/differ.service.ts
```diff
@@ -48,4 +49,96 @@ export class DifferService {
       .replace(/#TITLE#/g, titleText)
       .replace(/#DIFF_HTML#/, diffHTML);
   }
+
```
mheusgit, line 52: I like this approach.  Makes it easy to add future stat checks... we will we be doing soon.

src/notification/notification.service.ts
```diff
@@ -103,6 +106,23 @@ export default class SlackService implements NotificationService {
...
+    let endpointComponentsStatsBlock: Block | undefined = undefined;
+    if (stats !== undefined) {
+      endpointComponentsStatsBlock = {
+        type: 'context',
+        elements: [
+          {
+            type: 'mrkdwn',
+            text: `*Endpoints:*\n Added: ${stats.endpoints.added} / Removed: ${stats.endpoints.removed}`,
+          },
+          {
```
mheusgit, line 118: Small change.  If both add/removed are zero we should not show the stats

mheusgit (approved):
Everything appears to be in order

mheusgit (approved):
Still good

PR #217, opened 2022-10-21
API-21082/  doc server  oas consistent display ordering

mheusgit:
- Changed OAS.spec & OAS.transformed_spec columns to JSON rather than JSONB.  This prevents issues with OAS iterations getting their keys sorted by JSONB's order.  Grants full key order to provider
- Added JsonLegacyType override the default JsonType used in Mikro-ORM.  Forces ORM to use 'json' column type not the default 'jsonb'.
- Removed  @BeforeUpdate() & @BeforeCreate() due to them preventing Mikro-ORM from properly detecting OAS deltas
- Removed sortRecord() logic from oas.service.  SF OAS now follows whatever is given by provider


christineyost (approved):
Looks good to me!! Nice to see some sorting code removed! Tried it out locally and saw that original ordering is maintained.

mheusgit:
Merge delayed until most recent MR is completed.

stevelong00 (commented):
Looks good to me!

PR #219, opened 2022-10-24
API-18944/  Add new fields to the /metadata endpoint in docserver

mheusgit:
- Updated behavior for the "/docserver/apis/{{apiName}}/metadata", "/docserver/portal-store/metadata", & "/docserver/portal-store/metadata-all" endpoints to include new API & version field data


stevelong00 (commented):
Looks good, everything tests good on my machine.

PR #220, opened 2022-10-24
API-20570/ Delete API fails when an APIVersion has LOAST results

mheusgit:
- Corrected outdated model population during API deletion


christineyost (approved):
Looks good!

ediiotero (approved):
Looks good

PR #224, opened 2022-11-05
API-21184/ Provider Dashboard: change behavior of Run Test button

mheusgit:
- Updated Loast manager to dynamically pass suite IDs to LOAST tool


christineyost (approved):
Looks good

src/loast/loast.manager.ts
```diff
@@ -121,15 +122,6 @@ class LoastManager {
         //  Prevents token secured APIs from performing automatic ExampleGroup testing
       }
 
-      if (suiteIds) {
```
dsteeve, line 124: seems like this would be good to leave in so we know the suiteIds are coming thru.

mheusgit: They would be part of logged request.  If we change to including suite ID that a API version supports then extra logging would have to be added.

PR #226, opened 2022-11-09
API-21338/ Make securityApikey optional on APIVersion schema

mheusgit:
- Updated DocServer schema to allow securityApikey to be optional.  If not provided null saved to DB


stevelong00 (approved):
Looks good!

dsteeve (approved):
looks good!

PR #227, opened 2022-11-10
API-21437/ Fix loast.filter.ts eslint disables

mheusgit:
- Updated logic to avoid direct reference to Mikro-ORM Collections in none test code


christineyost (approved):
Looks good!

PR #229, opened 2022-11-15
API-15999/fix for migrations in QA

ediiotero:
Deleted migration files from the previous PR and ran another migration so all changes are in one file.


mheusgit:
I expect Migration20221103160003.ts would have to be removed

dsteeve (changes requested):
I think we need to copy over the major version data to version.

ediiotero:
Team went with a revert to fix this

PR #234, opened 2022-12-07
API-21174/  security apikey and security oauth validation

mheusgit:
- Updated DocServer schema to make securityApikey required if security array includes apiKey.  Similar change for securityOauth and oauth .


ediiotero (approved):
Looks good

PR #237, opened 2022-12-12
API-22364/ Docserver: Fix OASDiff URL to include correct path

mheusgit:
- Next/previous iteration URLs will now always include a 'v' to ensure they match with majorVersion to version revamp
- Change is intended to support backward compatibility for older OASDiffs


dsteeve (approved):
lgtm!

PR #238, opened 2022-12-14
API-22339/ Update Docserver Loast integration - Loast version package check

mheusgit:
- Switched to dynamic importing for loast package.json


dsteeve (approved):
👍

PR #239, opened 2022-12-22
Api 22214/Handling APIVersion updates

ediiotero:
Story: [API-22214](https://vajira.max.gov/browse/API-22214)

Added a new apiVersionPUTSchema and logic to prevent changing majorVersion or fhirVersion when updating an API version


src/api/version/apiVersion.schema.ts
```diff
@@ -72,3 +72,9 @@ export const apiVersionLookupSchema = Joi.object()
     version: Joi.string().required(),
   })
   .options({ abortEarly: false });
+
+export const apiVersionPUTSchema = apiVersionMutateSchema
```
mheusgit, line 76: Haven't seen this done before.  This approach should help reduce number of lines in other schemas in the near future.

mheusgit: Did a quick check.  The other schemas wont benefit from this as much as I'd hoped.

mheusgit:
Pulled this logic down for testing.  I noticed it failed to detect version updates if I moved from majorVersion "1" to "0"

src/api/version/apiVersion.service.ts
```diff
@@ -117,7 +117,32 @@ class APIVersionService {
   public async updateAPIVersion(
     model: APIVersionModel,
     updates: Optional<APIVersion>,
+    version: string,
   ): Promise<APIVersionModel> {
+    if (updates.majorVersion || updates.fhirVersion) {
+      const majorVersion = updates.majorVersion
```
mheusgit, line 123: I suspect "updates.majorVersion ?" if evaluating to false when majorVersion is "0" but the intent was to check if it was undefined

src/api/version/apiVersion.service.ts
```diff
@@ -117,7 +117,32 @@ class APIVersionService {
   public async updateAPIVersion(
     model: APIVersionModel,
     updates: Optional<APIVersion>,
+    version: string,
   ): Promise<APIVersionModel> {
+    if (updates.majorVersion?.toString() || updates.fhirVersion) {
```
dsteeve, line 122: would it make sense to add a test to try to change the api version from 1 to 0 to make sure this code block is exercised appropriately?

dsteeve (commented):
might want to add additional testing to cover the 0 case Marc mentions.

mheusgit (approved):
Looks good.  Requested changes appear to be in place

stevelong00 (commented):
Looks good!

PR #242, opened 2023-01-11
API-21561/  docserver  oas refresh triggers loast

mheusgit:
- Updated OAS refresh logic to trigger Loast oas-ruleset testing when a change is detected
- Moved higher level functions out of OASService into new OASManager.  Update included moving refreshAllOAS, refreshOAS, batchOAScleanup, and findOASAndUpload.  Key change is OASService no longer uses portal store or notifications services
- retransformOAS and refreshOAS received some cleanup to kill off duplicate code for OAS transformations.  Higher level operations of retransformOAS moved to OASManager
- OASService .resolveDelta & OASService.getLatestOASModel moved from private to public access

Followup changes:
- Fetching values from AWS Parameter Store has been commented out until positive test suite is part of test automation


test/oas/oas.manager.integration.ts
```diff
@@ -0,0 +1,227 @@
...
+  });
+
+  afterAll(async () => {
+    server.close();
+    await orm.close(true);
+    process.env = PREV_ENV;
+    configRefresh();
+  });
+
+  describe('refreshAllOAS', () => {
```
christineyost, line 124: Surprised we don't have a test for batchOASCleanup in here. Would you mind adding one or writing a ticket for it?

test/oas/oas.manager.test.ts
```diff
@@ -0,0 +1,521 @@
...
+    security: [{ oauth: [] }],
+  } as unknown as OAS;
+  const sfoas = oas;
+
+  let apiModel: APIModel;
+  let apiVersionModel: APIVersionModel;
+  let oasModel: OASModel;
+  let oasTransformer: OASTransformer;
+  let oasService: OASService;
+  let service: OASManager;
```
christineyost, line 45: Might be more clear to name this variable `manager`

src/oas/oas.service.ts
```diff
@@ -11,87 +11,26 @@ import OASModel from './oas.model';
 import StatusEnum from '../status/status.enum';
 import crypto from 'crypto';
 import OASTransformer from './transformer/oas.transformer';
-import SwaggerUITransformer from './transformer/oas.swaggerUI.transformer';
-import PortalStoreService from '../portalStore/portalStore.service';
-import { NotificationService } from '../notification/notification.interface';
 import { DifferService } from '../differ/differ.service';
```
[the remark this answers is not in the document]
mheusgit: Yeah it no longer belongs here but I left it alone in last commit.  I'll give it another look.

src/oas/oas.manager.ts
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    return newModel;
+  }
+
+  private async populateLoastResults(
+    apiVersion: APIVersionModel,
+  ): Promise<void> {
+    // Refreshes only trigger oas-ruleset tests & supporting single environment tests at the moment
+    const targetEnv = cfg.LOAST_DEFAULT_ENV;
```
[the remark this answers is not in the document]
mheusgit: Team agreed to not call AWS Param Store for the time being.

src/oas/oas.service.ts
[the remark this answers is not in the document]
mheusgit: OASService is now strictly reserved for OAS add, update, delete, and transform logic.

test/oas/oas.manager.test.ts
mheusgit: Updated

test/oas/oas.manager.integration.ts
mheusgit: Lets save it for a cleanup ticket

src/oas/oas.manager.ts
```diff
@@ -119,13 +120,38 @@ class OASManager {
...
+
+    return delta;
+  }
+
   private async populateLoastResults(
     apiVersion: APIVersionModel,
   ): Promise<void> {
     // Refreshes only trigger oas-ruleset tests & supporting single environment tests at the moment
-    const targetEnv = cfg.LOAST_DEFAULT_ENV;
+    // const targetEnv = cfg.LOAST_DEFAULT_ENV; Add back when positive test suite automated
```
christineyost, line 152: Not sure I understand what the LOAST_DEFAULT_ENV param is for. I think our targetEnv should always be the environment we're currently in, except for prod we should use sandbox because we don't want to run tests in prod. For example it wouldn't make sense to run tests against the dev oas in sandbox.

I think some of this design might have to be rethought. For example, even when LOAST positive is automated, I think we'd only want to run it in the sandbox environment, because we won't know if the same examples would work in other environments. I think we need to track what LOAST suites should be run for each APIVersion in the LOAST config. That way we could automate the positive suite for APIVersions that we know it works for as well.

I'd be fine with this going in like this, but I think we need a spike to think this through and maybe do some cleanup. cc @dsteeve

mheusgit: Team agreed awhile back the positive test would be run against just sandbox.  I recall its even mentioned in the confluence doc.  Let me check

mheusgit: Yes.  Its the "[the] actual API environment being tested against (not the environment LOAST is run on).  Should just be "sandbox" for the foreseeable future.  Might include "production" or "staging" far down the road"

dsteeve: great discussion and points.  I think some of this was designed prior to implementing the oas-ruleset test and now we have some new functionality to take into consideration.  New functionality being that oas-ruleset would run in every environment(qa,staging,sandbox) but positive test would only run in sandbox unless some APIs have example groups that can be used in any environment.   Instead of hardcoding all that business logic in our code, I would like to see a more data-driven approach.   In the future we might have like a authenticationTest that would run in every env.   @christineyost can you create a spike ticket to review this design?

test/oas/oas.manager.integration.ts
dsteeve: can you write a ticket for this @mheusgit ?

mheusgit: sure

christineyost: [API-23417](https://vajira.max.gov/browse/API-23417)

mheusgit:
PR should be good to go

christineyost (approved):
Looks good!

PR #245, opened 2023-01-25
API-18232/ enable sql logging by default if LOG_LEVEL=silly

jeremyhwu:


src/app.ts
```diff
@@ -50,6 +50,10 @@ export default async function configureApp(
 
   await migrate();
   const injector = await ServiceInjector.initialize(orm, notificationService);
+  // Enable SQL logging by default when log level is 'silly'
+  if (cfg.LOG_LEVEL === 'silly') {
+    injector.adminService.setLogLevel(injector.adminService.getLogLevel());
```
mheusgit, line 55: injector.adminService.setLogLevel({logLevel: 'silly'}); Seems less confusing

mheusgit (approved):
Should be good

PR #246, opened 2023-02-03
API-23305/  remove loast config model authentication type field

mheusgit:
- Removed authentication type field from loast config

Changes in PR https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/242 need to be merged first


src/loast/loast.manager.ts
```diff
@@ -99,25 +99,31 @@ class LoastManager {
 
       logger.info(testOptions);
 
-      const apiCredentialName = cfg.LOAST_CREDENTIAL_PATTERN.replace(
+      /* 
+        const security = apiVersion.security;
+        const credentialType = security ? security[0] : '';
```
christineyost, line 104: If security is set credentials should be fetched for each security type. LOAST can already handle having both an apikey and a token passed in. We shouldn't have to rely on only the first item

src/loast/loast.manager.ts
```diff
@@ -81,7 +81,7 @@ class LoastManager {
       await this.entityManager.populate(apiVersion, ['loastConfig']);
       const config = apiVersion.loastConfig;
 
-      if (!config || !config.authenticationType || !config.oasPath) {
+      if (!config || !config.oasPath) {
         logger.info(
```
dsteeve, line 85: while we are in this code, this should probably be an error and not info to help our future selves in the event there is a prod problem and logging is only at error level.

src/loast/loast.manager.ts
dsteeve: another option is to just remove the commented code for now

christineyost: Agree, I do think this code might change quite a bit for oAuth tokens depending on how we fetch oAuth tokens

mheusgit: See comment below: "Should be moved to a loop based on security array"

src/loast/loast.manager.ts
mheusgit: Updated

dsteeve (approved):

PR #248, opened 2023-02-08
API-18323 - Set query debug mode based on startup log level

christineyost:
This PR is for [API-18323](https://vajira.max.gov/browse/API-18232). It sets MikroORM's logger debug mode based on the startup log level.


mheusgit (approved):
Looks good

src/admin/admin.service.ts
```diff
@@ -27,17 +27,22 @@ class AdminService {
...
-      }
+      this.setQueryLogging();
     }
 
     return logChanged;
   }
+
+  // Enables/disables query logging based on log level
+  public setQueryLogging(): void {
+    if (logger.level === 'silly') {
```
dsteeve, line 38: one small improvement is to try and avoid hardcoding "silly" here.  
I think that logger.isSillyEnabled() might work?  didn't try it.

dsteeve (approved):
thanks!

PR #249, opened 2023-02-08
API-23432/Update APIVersion Model security

ediiotero:
Story: [API-23432](https://vajira.max.gov/browse/API-23432)

Update APIVersion model to require security field.
Update Unit and Integration tests.


src/api/version/apiVersion.model.ts
```diff
@@ -115,7 +115,7 @@ class APIVersionModel extends BaseEntity {
   @Property({ fieldName: 'security', nullable: true })
   securityCSV?: string;
```
christineyost, line 116: I think this can be marked as required also since this is the property that gets stored in the database

src/api/version/apiVersion.interface.ts
```diff
@@ -30,7 +30,7 @@ export interface APIVersion {
   status: APIVersionStatusEnum;
   refresh: boolean;
   intendedUse?: APIVersionIntendedUseEnum;
-  security?: APIVersionSecurityEnum[];
+  security: APIVersionSecurityEnum[];
```
christineyost, line 33: The security field in the APIVersionResponseBody below can also be marked required since it will always be returned

src/api/version/apiVersion.model.ts
christineyost: This will also require a database migration

christineyost (commented):
I think we can also mark the `securityCSV` field in APIVersionModel as required since that's where the security values are stored in the database. This will also require a migration to be created.

It also looks like there are some more places in tests where the APIVersionModel constructor needs to be updated. Running `npm run build:tests` will give a list of places where the security parameter needs to be added

src/api/version/apiVersion.model.ts
```diff
@@ -171,10 +171,11 @@ class APIVersionModel extends BaseEntity {
     this.internalOnly = internalOnly;
     this.status = status;
     this.refresh = refresh;
-
+    this.securityCSV = security.join();
```
christineyost, line 174: Is this needed here? It looks like this will get set when `createCSVs` is called at the very end of this constructor

test/api/version/apiVersion.controller.integration.ts
```diff
@@ -53,7 +53,6 @@ describe('APIVersions Controller', () => {
     renderConfig: {},
     loastConfig: {},
     intendedUse: APIVersionIntendedUseEnum.INTERNAL,
-    security: [APIVersionSecurityEnum.APIKEY],
```
christineyost, line 56: I think this should stay - we'd still expect this to be returned

christineyost (commented):
Looks good, just had a couple questions

christineyost (approved):
Looks good!

Before merging it might be nice to make sure that the security field in the database is already set for every APIVersion or we'll run into problems, if that hasn't been done already

PR #250, opened 2023-02-10
Api 24211 fix build test errors

dsteeve:
This is to fix the 6 errors found when doing "npm run build:tests" after API-21561 was merged in.
This was difficult to see until API-22334 was merged in to fix the other noise.


stevelong00 (dismissed):
Confirmed that the build completes without error and all tests pass.  Looks good!

mheusgit (approved):
Looks good

PR #252, opened 2023-02-17
API-24174 - update LOAST version

christineyost:
This PR is for [API-24174](https://vajira.max.gov/browse/API-24174). It updates the LOAST version in Docserver to include the bug fix in LOAST for the oas-ruleset "Cannot read properties of null" error


mheusgit (approved):
Change is as expected.

PR #253, opened 2023-02-18
API-22809/ DocServer: API versions last updated should include dependants updates

mheusgit:
- Updated filtering for renderConfig, loastConfig, and gatewayConfig to stop filtering out the updatedAt property


src/api/gatewayConfig/gatewayConfig.interface.ts
```diff
@@ -10,4 +10,5 @@ export interface GatewayConfigResponse {
   routeId?: string;
   removePath?: string;
   exposedRoute?: string;
+  updatedAt: Date;
```
christineyost, line 13: Should this be optional like in the other response interfaces?

mheusgit: It should be.

mheusgit: Updated

christineyost (approved):
Looks good to me!

dsteeve (approved):
lgtm!

PR #254, opened 2023-02-22
Api 23299/APIVersion Model remove Security Fields

ediiotero:
Story: [API-23299](https://vajira.max.gov/browse/API-23299)

Remove securityType and securityModel from APIVersionModel


src/oas/transformer/oas.transformer.ts
```diff
@@ -31,10 +32,13 @@ class OASTransformer {
...
     if (
-      apiVersion.securityModel.type === SecurityTypeEnum.API_KEY &&
+      apiVersion.security.includes(APIVersionSecurityEnum.APIKEY) &&
+      apiVersion.security.length === 1 &&
       this.shouldUpdateSecurity(clonedSpec.components?.securitySchemes)
     ) {
-      const securitySchema = apiVersion.securityModel.getSchema();
+      const securitySchema = BaseSecurity.constructSecurity(
+        apiVersion.security[0],
+      ).getSchema();
```
christineyost, line 41: Instead of using this constructSecurity method here I think we could simplify this logic by storing the apikey security scheme json blob in a resource file.

The security scheme json blob is:
```
{
  APIVersionSecurityEnum.APIKEY: {
      in: OASSecuritySchemeIn.HEADER,
      name: APIVersionSecurityEnum.APIKEY,
      type: OASSecuritySchemeType.APIKEY,
    }
}
```

This could be stored in a securitySchemes.ts file along side and much like [gatewayResponses.ts](https://github.com/department-of-veterans-affairs/lighthouse-docserver/blob/main/src/oas/transformer/resources/gatewayResponses.ts)

Then this json blob could be sent directly into the updateComponentsElement method call below.

With this change we should be able to remove all the files in `src/api/security`

test/api/security/base.security.test.ts
```diff
@@ -1,12 +1,12 @@
 import BaseSecurity from '../../../src/api/security/base.security';
-import SecurityTypeEnum from '../../../src/api/security/securityType.enum';
+import { APIVersionSecurityEnum } from '../../../src/api/version/apiVersion.enum';
 
 describe('BaseSecurity', () => {
```
christineyost, line 4: This test file should be able to be deleted

test/oas/transformer/oas.security.transformer.test.ts
```diff
@@ -1,31 +1,35 @@
...
 import OASSecurityTransformer from '../../../src/oas/transformer/oas.security.transformer';
+import { APIVersionSecurityEnum } from '../../../src/api/version/apiVersion.enum';
 
 describe('OASSecurityTransformer', () => {
   describe('getSecurityElement', () => {
     it('returns empty scopes for api key', () => {
       const security = BaseSecurity.constructSecurity(
-        SecurityTypeEnum.API_KEY,
+        APIVersionSecurityEnum.APIKEY,
       ).getSchema();
```
christineyost, line 10: Instead of constructing this scheme here the json blob mentioned above could be referenced

test/oas/transformer/oas.security.transformer.test.ts
```diff
@@ -1,31 +1,35 @@
...
       const securityRequirement =
         OASSecurityTransformer.getSecurityElement(security);
       expect(securityRequirement.length).toEqual(1);
-      expect(securityRequirement[0][SecurityTypeEnum.BEARER]).toBeTruthy();
-      expect(securityRequirement[0][SecurityTypeEnum.BEARER].length).toEqual(0);
+      expect(securityRequirement[0][APIVersionSecurityEnum.OAUTH]).toBeTruthy();
+      expect(
+        securityRequirement[0][APIVersionSecurityEnum.OAUTH].length,
+      ).toEqual(0);
     });
```
christineyost, line 33: This test can be removed since we don't set this for oauth APIs

src/api/security/apiKey.security.ts
```diff
@@ -5,10 +5,10 @@ import {
   OASSecuritySchemeType,
 } from '../../oas/oas.interface';
 import BaseSecurity from './base.security';
-import SecurityTypeEnum from './securityType.enum';
+import { APIVersionSecurityEnum } from '../version/apiVersion.enum';
 
 class ApiKeySecurity implements BaseSecurity {
```
christineyost, line 10: This file should be able to be deleted

src/api/security/base.security.ts
```diff
@@ -1,18 +1,18 @@
 import { Dictionary, OASSecurityScheme } from '../../oas/oas.interface';
 import ApiKeySecurity from './apiKey.security';
 import BearerSecurity from './bearer.security';
-import SecurityTypeEnum from './securityType.enum';
+import { APIVersionSecurityEnum } from '../version/apiVersion.enum';
 
 abstract class BaseSecurity {
```
christineyost, line 6: This file should be able to be deleted

src/api/security/bearer.security.ts
```diff
@@ -4,10 +4,10 @@ import {
   OASSecuritySchemeType,
 } from '../../oas/oas.interface';
 import BaseSecurity from './base.security';
-import SecurityTypeEnum from './securityType.enum';
+import { APIVersionSecurityEnum } from '../version/apiVersion.enum';
 
 class BearerSecurity implements BaseSecurity {
```
christineyost, line 9: This file should be able to be deleted

src/oas/transformer/oas.paths.transformer.ts
```diff
@@ -140,9 +138,9 @@ class OASPathsTransformer {
   }
 
   private getGatewayResponses(
-    securityType: SecurityTypeEnum,
+    security: APIVersionSecurityEnum[],
   ): Dictionary<OASResponse> {
-    if (securityType === SecurityTypeEnum.API_KEY) {
+    if (security.includes(APIVersionSecurityEnum.APIKEY)) {
```
christineyost, line 143: I think this is a good approach for determining gateway responses for now. We could potentially try to determine the security of each operation, but that would complicate the logic and I'm not sure what we would do if we can't determine the security. I think this is a good approach especially for now because we don't have any APIs that use both oAuth and APIkey, but I think we should run it by the team, maybe in standup tomorrow?

christineyost: We agreed to use this logic for now. Created [API-24728](https://vajira.max.gov/browse/API-24728) to add more complex logic later

christineyost:
Sorry I'd missed this before... the data dictionary in the README should also be updated to remove the security_type field from APIVersion. The APIVersion security field description should also be updated to include that the field is used for OAS transformation

src/oas/transformer/resources/securitySchemes.ts
```diff
@@ -0,0 +1,23 @@
...
+  },
+};
+
+export const oauthSecurityScheme: Dictionary<OASSecurityScheme> = {
+  [APIVersionSecurityEnum.OAUTH]: {
+    type: OASSecuritySchemeType.HTTP,
+    scheme: 'bearer',
+    bearerFormat: 'JWT',
+  },
+};
```
christineyost, line 23: This scheme can be removed - it's never used

test/oas/transformer/oas.security.transformer.test.ts
```diff
@@ -1,31 +1,29 @@
...
-        SecurityTypeEnum.BEARER,
-      ).getSchema();
       const securityRequirement =
-        OASSecurityTransformer.getSecurityElement(security);
+        OASSecurityTransformer.getSecurityElement(oauthSecurityScheme);
       expect(securityRequirement.length).toEqual(1);
-      expect(securityRequirement[0][SecurityTypeEnum.BEARER]).toBeTruthy();
-      expect(securityRequirement[0][SecurityTypeEnum.BEARER].length).toEqual(0);
+      expect(securityRequirement[0][APIVersionSecurityEnum.OAUTH]).toBeTruthy();
     });
```
christineyost, line 27: This test can be deleted

christineyost (commented):
There's one more file that should be able to be removed, src/api/security/securityType.enum.ts

src/oas/transformer/oas.transformer.ts
```diff
@@ -31,17 +32,17 @@ class OASTransformer {
     );
 
     if (
-      apiVersion.securityModel.type === SecurityTypeEnum.API_KEY &&
+      apiVersion.security.includes(APIVersionSecurityEnum.APIKEY) &&
```
mheusgit, line 35: Pretty sure this will fail when security is not provided.  Data model shows that field is optional.

mheusgit:
Looked over the changes and didn't see anything that stood out.

migrations/Migration20230221001529.ts
```diff
@@ -0,0 +1,13 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20230221001529 extends Migration {
```
mheusgit, line 3: No changes for .snapshot-docserver.json?

ediiotero: I did run the migration not sure why that wasn't removed. I ran it again seems to have picked it up this time.

migrations/Migration20230309190435.ts
```diff
@@ -0,0 +1,13 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20230309190435 extends Migration {
```
mheusgit, line 3: The old migration file needs removal now

dsteeve: @mheusgit - not sure I understand, we are still dropping the security_type field, correct?

mheusgit: Prior to the https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/254/commits/3cc65df1ccd7037a2119c1142a0fd2bb60cca929 commit there were two migrations performing the same DB updates

dsteeve: 👍

dsteeve (commented):
looks good to me, but will merge in the loast-config-suites change first, then bring this branch up to date after that and re-test it

christineyost (approved):
Looks good to me! Changes should be merged along with the Provider Dashboard change to remove the security type field on the APIVersion form

mheusgit (approved):
Looks good

PR #256, opened 2023-02-27
API-24549 add diff detection for extra long text fields

jeremyhwu:


mheusgit (approved):
Not much to say about change other than it follows JIRA request.  

This shouldn't affect upcoming MR due to the nature of how oasdiff.js is deployed but hold off on merge until tomorrow.

PR #257, opened 2023-02-28
API-21581 - Add LoastSuite model, service, and controller

christineyost:
This PR is for [API-21581](https://vajira.max.gov/browse/API-21581). It adds a LoastSuite model, service, and controller to Docserver. LoastSuites can be created, read, updated, and deleted using Docserver endpoints.


dsteeve (commented):
small readme fix is the only thing I found.  looks great!

dsteeve (approved):
looks good!

PR #259, opened 2023-03-08
API-21581 - Add LoastSuite/LoastConfig relationship

christineyost:
This PR is for [API-21581](https://vajira.max.gov/browse/API-21581). It creates the many to many relationship between LoastSuite and LoastConfig. It also creates an APIVersionManager since when creating and updating an APIVersion that includes a LoastConfig, LoastSuites will need to be fetched. Many test needed to be updated as well.

One remaining question is when a LoastSuite is deleted, should any LoastConfig that only had that LoastSuite also be deleted?


mheusgit:
Between this PR and https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/254 Jest test in CI/DI will likely fail due to the number of migration files.

dsteeve (approved):
I tried this out locally, looks good to me. 
readme addition looks good as well.

PR #260, opened 2023-03-08
API-21562/  docserver  update slack messaging to include loast statistics

mheusgit:
- Updated OAS refresh messaging to include LOAST statistics and URL to LOAST when meaningful changes were made to OAS
- Statistics interface name changed to OASStatistics now that there is another type of stats being returned
- Updates include name change for populateLoastResults() since it now performs some statistics handling


src/loast/loast.manager.ts
```diff
@@ -160,6 +168,74 @@ class LoastManager {
...
+  public async getLatestLoastStatistics(
+    apiVersion: APIVersionModel,
+  ): Promise<LOASTStatistics | undefined> {
+    let loastStats: LOASTStatistics | undefined = undefined;
+    const requestType = getContext('requestType') as RequestTypeEnum;
+
+    // Only calculate when called by automatic request
+    if (RequestTypeEnum.SYSTEM !== requestType) {
+      logger.info(`SKIPPED: LOAST stats calculation`);
+      return await loastStats;
```
christineyost, line 187:
```suggestion
      return loastStats;
```

src/loast/loast.manager.ts
```diff
@@ -160,6 +168,74 @@ class LoastManager {
...
+      if (loastRequests.length > 0) {
+        const latestRequest = loastRequests[0];
+        let matchingRequest: LoastRequest | undefined = undefined;
+
+        for (let x = 1; x < loastRequests.length; x++) {
+          if (latestRequest.suites === loastRequests[x].suites) {
+            matchingRequest = loastRequests[x];
+            break;
+          }
+        }
```
christineyost, line 206:
```suggestion
        const matchingRequest = loastRequests.find(request => latestRequest.suites === request.suites);
```

src/loast/loast.manager.ts
```diff
@@ -160,6 +168,74 @@ class LoastManager {
...
+            `LOAST stats only found requestId: ${latestRequest.requestId}`,
+          );
+        }
+
+        loastStats = {
+          failures: Number(latestRequest.failures),
+          failuresDiffs: failuresDiffs,
+          warnings: Number(latestRequest.warnings),
+          warningDiffs: warningDiffs,
+        } as LOASTStatistics;
```
christineyost, line 231:
```suggestion
        };
```

src/loast/loast.service.ts
```diff
@@ -57,6 +58,29 @@ class LoastService {
...
+      .createQueryBuilder(LoastResultModel, 'l')
+      .select(select)
+      .where({ api_version_id: apiVersion.id })
+      .groupBy('l.request_id')
+      .orderBy({ 'Max(l.run_datetime)': 'desc' })
+      .execute()) as LoastRequest[];
+
+    return loastRequests;
+  }
 }
```
christineyost, line 84: Might be worth adding an integration test that tests this logic using the database, especially since we don't use the query builder anywhere else

dsteeve: I will add a "if time" comment here too.  If it is doable, I've seen in other codebases when a complicated query is being built, that there is a logger.debug(sqlstatement) logged -- so if there are issues, debug can be turned on and we can see the actual query being run. However, we may already have that by turning on orm debugging?

src/loast/loast.manager.ts
mheusgit: We need to skip the first request as it will always match with latestRequest.

src/loast/interfaces/loast.common.interface.ts
```diff
@@ -51,3 +51,11 @@ export function isLoastEndpointSummary(
 ): input is LoastEndpointSummary {
   return (input as LoastEndpointSummary).totalPass !== undefined;
 }
+
+// Used to compare LOAST test runs.  Diffs will be negative if new test reports fewer issues
+export interface LOASTStatistics {
```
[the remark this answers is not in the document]
mheusgit: Updated

src/loast/loast.manager.ts
mheusgit: Typo removed

src/loast/loast.service.ts
mheusgit: We have Mikro-ORM level logging of all queries.  You just need to turn on finest level logging to see them.

src/notification/notification.service.ts
```diff
@@ -105,36 +107,86 @@ export default class SlackService implements NotificationService {
     const fallbackTime = oas.updatedAt.toUTCString(); // Needed if Slack client fails to render <!date>
     const viewOASURL = `${urlBase}/internal/docs/${apiVersion.api.name}/${apiVersion.version}/openapi.json`;
     const viewChangesURL = `${urlBase}/internal/docs/oasdiff/${apiVersion.api.name}_${apiVersion.version}_i${prevIteration}_i${iteration}.html${queryStr}`;
+    const viewLOASTURL = cfg.LOAST_URL_PATTERN.replace(
+      `{API_NAME}`,
+      apiVersion.api.name,
+    ).replace(`{VERSION}`, `${apiVersion.version}`);
+    let showLOASTLink = false;
```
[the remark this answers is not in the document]
mheusgit: Naming was updated

src/loast/loast.manager.ts
christineyost: What about something like
```
const latestRequest = loastRequests[0];
const remainingRequests = loastRequests.slice(1);

const matchingRequest = remainingRequests.find(request => latestRequest.suites === request.suites);
```

mheusgit: Logic changed

christineyost (approved):
Looks good!

PR #264, opened 2023-03-23
API-24660 (Fix): Use path to define file path for oasdiff files

jeremyhwu:
Fixing S3 uploads for oasdiff files by adding a relative path. Currently, logs in aws are showing an `ENOENT` error due to the path being wrong.


mheusgit (approved):
Looks good

PR #267, opened 2023-03-24
API-25139 - Update LoastManager to use LoastSuites from LoastConfig

christineyost:
This PR is for [API-25139](https://vajira.max.gov/browse/API-25139). It updates LoastManager to make use of LoastSuites in LoastConfig. It required some refactoring to return 400 statuses with relevant error messages when a request is made to `POST /loast` with a bad request body - requesting an APIVersion that does not have LoastConfig set, requesting a suiteId that is not set on LoastConfig, or requesting an APIVersion with a LoastConfig that doesn't have any LoastSuites set.

Removed the targetEnv variable for fetching credentials. Now target env will correspond to the Docserver environment - Docserver QA/dev = dev, Docserver staging = staging, Docserver prod = sandbox.

Also removed SuiteIdEnum so that now as new suites are added we will only need to add new data to Docserver, hopefully not make code changes.

Once these changes are in we should be able to set up the `positive` suite for APIs that support it, only in the Docserver PROD (Lighthouse SANDBOX) environment.


src/loast/loast.manager.ts
```diff
@@ -233,6 +233,100 @@ class LoastManager {
...
+
+    return suites;
+  }
+
+  private async fetchAPIKeyCredential(
+    apiName: string,
+    version: string,
+  ): Promise<string> {
+    const credentialType = APIVersionSecurityEnum.APIKEY;
+    const env = cfg.IS_PRODUCTION ? 'sandbox' : cfg.ENV_NAME;
```
mheusgit, line 310: I see the target environment config was removed.

Previous approach was based on a expectation a environment DocServer might need to perform 'positive' testing against multiple API environments.  Controller could potentially pass the value down as parameter later on.

This change means each DocServer environment will be tied to a single API environment.  QA/DEV DocServer positive tests run fail as these tests are supposed to be against sandbox.  STAGING DocServer positive test runs will also fail.

christineyost: I think this approach of tying Docserver LOAST runs to one API environment is better. In the positive suite, in QA/DEV for example, we use the QA/DEV OAS to create API requests. It might not make sense to execute those requests against the SANDBOX environment because the SANDBOX environment API might not match. For example If an API team is working on adding a new endpoint, that endpoint might be present in the QA/DEV OAS, but it might not be immediately available in the SANDBOX environment until testing is finished.

I think the positive suite should only be run in the Docserver PROD/Lighthouse SANDBOX environment

mheusgit: Yes.  Now the positive test suite can only be triggered on DocServer PROD.

src/loast/loast.manager.ts
```diff
@@ -233,6 +233,100 @@ class LoastManager {
 
     return loastStats;
   }
+
+  public getLoastConfigOrFail(apiVersion: APIVersionModel): LoastConfigModel {
```
mheusgit, line 237: Method is unused.  Belongs in service

mheusgit (approved):

PR #273, opened 2023-04-13
[API-25396] Add optional query parameter to show s3 status for api versions

jeremyhwu:
Adding an optional query parameter `is_published` to show the status of an API Version according to the S3 metadata.json. Also added a `get` function to the S3Client since it hasn't been implemented yet.


src/api/version/apiVersion.controller.ts
```diff
@@ -113,8 +118,35 @@ class APIVersionController {
...
+
+      // Optional query parameter that shows the S3 status of each API version
+      const { include_published = 'false' } = request.query;
+      if (include_published === 'true') {
+        const s3Response = await this.s3Service.get(
+          `${api.name}/metadata.json`,
+        );
+
+        const s3ResponseString = s3Response.toString();
+        const metadata = JSON.parse(s3ResponseString) as Metadata;
```
christineyost, line 133: Might be nice to add a function to the PortalStoreService that gets metadata by API name. If we ever need to switch from S3 to a different datastore that way our changes will all be in the PortalStoreService instead of throughout the code

src/utils/s3.service.ts
```diff
@@ -130,6 +132,33 @@ class S3Service {
...
+
+      return new Promise<Buffer>((resolve, reject) => {
+        stream.on('data', (chunk) => chunks.push(chunk));
+        stream.once('error', (err) => reject(err));
+        stream.once('end', () => resolve(Buffer.concat(chunks)));
+      });
+    } catch (err) {
+      logger.error(err);
+      throw new ServerError(err as Error, `Unable to get file in s3 ${key}`);
+    }
```
christineyost, line 159: Need to handle the file not existing. If the metadata file does not exist then none of the versions are published

src/utils/s3.service.ts
```diff
@@ -130,6 +132,33 @@ class S3Service {
...
+    try {
+      const response = await this._client.send(getObjectCommand);
+      const stream = response.Body as Readable;
+      const chunks: Buffer[] = [];
+
+      return new Promise<Buffer>((resolve, reject) => {
+        stream.on('data', (chunk) => chunks.push(chunk));
+        stream.once('error', (err) => reject(err));
+        stream.once('end', () => resolve(Buffer.concat(chunks)));
+      });
```
christineyost, line 155: This might be able to be simplified using `sdkStreamMixin(response.Body).transformToString()`

`sdkStreamMixin` comes from `@aws-sdk/util-stream-node`

src/utils/s3.service.ts
jeremyhwu: This is now handled in the `getMetadata` function from `PortalStoreService`.

src/api/version/apiVersion.controller.ts
jeremyhwu: Added in latest commit

src/portalStore/portalStore.service.ts
```diff
@@ -70,6 +70,22 @@ class PortalStoreService {
...
+      const metadata = JSON.parse(s3Response) as Metadata;
+      return metadata;
+    } catch (err) {
+      // If the metadata file does not exist, none of the versions have been published
+      return {
+        meta: {
+          versions: [],
+        },
+      } as unknown as Metadata;
+    }
```
christineyost, line 87: I think this makes sense to me - if fetching metadata from S3 results in any kind of error, treat it as if no APIVersions are published for the API. Just want to double check with @dsteeve on this

dsteeve: I think it might make sense to add a static method to metadata.model.ts that returns a default empty Metadata with the "meta: { versions: [],}. instead of having it inline here.  Reason being is that this could be a bug over time when the interface changes - and having it in a method with the model will more likely get caught and changed by a developer if/when the interface changes.

I also think we can't assume that any error should do this, but hopefully we can only do this when it is a 404 error (nothing found) from the s3Service. It is possible that the network is down or AWS is down, then we want to do something different and maybe don't put the "publish" in there at all if it is a 500 error from AWS.

src/api/version/apiVersion.controller.ts
```diff
@@ -113,8 +117,30 @@ class APIVersionController {
       const apiVersions = await this.apiVersionService.getAPIVersionsForAPI(
         api,
       );
+      const filteredAPIVersionResponse = apiVersions.map(
+        filterAPIVersionResponse,
+      );
+
+      // Optional query parameter that shows the S3 status of each API version
+      const { include_published = 'false' } = request.query;
+      if (include_published === 'true') {
```
dsteeve, line 126: This feels like too much business logic at the controller level.  Typically I think the pattern has been for the service or manager to return a model and then the controller takes the model and builds the response.  

Since this controller already has the APIVersionManager, I would suggest that it get used and this logic get pushed into there.  
Perhaps a apiVersionManager.getAPIVersionsForAPI that takes the api, and this include_published as parameters. 

Then, it calls the apiVersionService and the portalStoreService coming back with the array of apiVersionModel filled in.  The apiVersionModel can then be used by the controller to create the response object.  

Unit testing might become easier as well with this approach

dsteeve (changes requested):

mheusgit (commented):
Did a review of current state.  The main item noticed was mentioned by dsteeve concerning working with PortalStore in the controller.  Prefer to use the manager class for that

src/api/version/apiVersion.manager.ts
```diff
@@ -99,6 +104,54 @@ class APIVersionManager {
...
+          apiVersion.published = true;
+        } else {
+          apiVersion.published = false;
+        }
+      }
+    }
+
+    return apiVersions;
+  }
+
```
christineyost, line 154: I'd suggest combining these two methods into one. I think we could do something like:

```
  public async getAPIVersionsForAPI(
    api: APIModel,
    include_published = false,
  ): Promise<APIVersionModel[]> {
    let apiVersions: APIVersionModel[];
    try {
      const versions = await api.versions.init({
        orderBy: { version: 'desc' },
        populate: ['loastConfig.suites'],
      });

      apiVersions = versions.getItems();
    } catch (err) {
      throw new ServerError(
        err as Error,
        `Getting versions for API ${api.name}`,
      );
    }

    if (include_published === true) {
      let metadata: Metadata;
      try {
        metadata = await this.portalStoreService.getMetadata(api);
      } catch (err) {
        return apiVersions;
      }

      const publishedVersions = metadata.meta.versions.map(
        (version) => version.version,
      );

      for (const apiVersion of apiVersions) {
        if (publishedVersions.includes(apiVersion.version)) {
          apiVersion.published = true;
        } else {
          apiVersion.published = false;
        }
      }
    }

    return apiVersions;
  }
```

Then in APIVersionController this getAPIVersionsForAPI function can be called instead of listAPIVersionsForAPI

test/fixtures/apiVersions.ts
```diff
@@ -9,6 +9,30 @@ import LoastSuiteModel from '../../src/loast/loastSuite/loastSuite.model';
...
+  { serviceId: 'one-ring', routeId: 'one-ring-api' },
+  'http://one-ring-oas',
+  1,
+  apiModel,
+  true,
+  APIVersionStatusEnum.CURRENT,
+  true,
+  [APIVersionSecurityEnum.APIKEY],
+  [],
+);
```
christineyost, line 34: It may be possible to use existing APIVersion fixtures instead of creating new ones. If you want one to be v0 and one to be v1 you could use `apiVersionKeyRemovePath` and `apiVersionAPIKeyAndRoute`

test/api/version/apiVersion.controller.test.ts
```diff
@@ -71,11 +90,44 @@ beforeEach(() => {
   );
 });
 
+it('returns a 200 and apiVersion data but does not show s3 published status', async () => {
```
christineyost, line 93:
```suggestion
it('APIVersionController.getAPIVersionsForAPI() returns a 200 and apiVersion data but does not show s3 published status', async () => {
```

Marc introduced this new pattern of not using `describe` statements and including "<class>.<method>()" at the beginning of each test description

test/api/version/apiVersion.controller.test.ts
```diff
@@ -71,11 +90,44 @@ beforeEach(() => {
...
+  );
+
+  expect(response.status).toHaveBeenCalledWith(200);
+  expect(response.json).toHaveBeenCalledWith([apiVersionKeyRemovePath]);
+  expect(response.json).not.toHaveBeenCalledWith([
+    { published: true, ...apiVersionKeyRemovePath },
+  ]);
+});
+
+it('returns a 200, apiVersion data, and includes s3 published status', async () => {
```
christineyost, line 110:
```suggestion
it('APIVersionController.getAPIVersionsForAPI() returns a 200, apiVersion data, and includes s3 published status', async () => {
```

test/api/version/apiVersion.manager.test.ts
```diff
@@ -138,3 +142,67 @@ it('APIVersionManager.updateAPIVersion() resolves gatewayconfig', async () => {
...
 });
+
+it('resolves gatewayconfig', async () => {
+  await apiVersionManager.updateAPIVersion(
+    apiVersionKeyRemovePath,
+    { gatewayConfig: { routeId: 'the-path' } },
+    'v0',
+  );
+  expect(gatewayService.resolveGatewayConfig).toHaveBeenCalled();
+});
```
christineyost, line 153: I think this is the same test as the one just above. It can be removed

test/api/version/apiVersion.manager.test.ts
```diff
@@ -138,3 +142,67 @@ it('APIVersionManager.updateAPIVersion() resolves gatewayconfig', async () => {
...
+
+describe('getAPIVersionsForAPI', () => {
```
christineyost, line 155: Similar to the APIVersionController test... the describe statement can be removed and add `APIVersionManager. getAPIVersionsForAPI()` to the beginning of each description in the it statements.

We're also trying to move away from having `beforeEach` functions, except at the root of the test file. Instead of having the before each here this mock can be added to the beginning of each of these tests that needs it

christineyost (approved):
Changes look good to me! Tested the query parameter locally as well as generating metadata and running LOAST via API batch

dsteeve (approved):
one small typo I noticed, but I brought it down, tested it.  Looks Good!

PR #274, opened 2023-04-14
API-23423/  Docserver  jest nesting cleanup

mheusgit:
- Moved DB related test setup into a new DB helper for consistent import.  Same approach was used for a new gateway helper and Node environment helper.  This removed hundreds of repeating lines of code.  
- Flattened API and API version test.  Moved a number of reusable parts to fixtures.  Also removed most before* after* hooks.  Exception was made for async beforeAll() needed for integration testing
- New fixtures for controller, managers, services, and associated assets created.  After switching to fixtures a couple hundred of lines of redundant code was removed.
- Moved jestSetEnvVars to a new globalSetup file.  Previously code was being run for each suite.
- Moved jest.config.js, jest.*.config.js, and tsconfig.test.json files into "/test" folder
- Added a performance tweak connected to MikroORM migrator() being run many times.  Now it will occur just once per test run
- Changed Jest timeout to 5 seconds


jeremyhwu (approved):

jeremyhwu (approved):

dsteeve (approved):
I think this moves us into a better pattern.

PR #275, opened 2023-04-21
API-25850/ DocServer: Correct Misnamed Gateway properties - initial changes

mheusgit:
- Updated naming for routeId and serviceId to better match how its used as routeName and serviceName.  Gateway IDs and the human readable "names" are now properly labeled.


mheusgit:
Closed PR and moved assets over to more recent branch https://github.com/department-of-veterans-affairs/lighthouse-docserver/pull/276

PR #281, opened 2023-05-02
API-26108/  doc server jenkins logging disable

mheusgit:
- Made small change so logs no longer appear during Jenkins testing
- Previous logic depended on IS_TEST_NODE (a.k.a. process.env.NODE_ENV) setting being configured but some test alter this


christineyost (approved):
Jenkins log is much more readable now!!! 🎉

PR #282, opened 2023-05-02
API-18595 - Update SchedulerController endpoint to handle the LOAST endpoint

christineyost:
This PR is for [API-18595](https://vajira.max.gov/browse/API-18595). It updates the SchedulerController endpoint to handle the POST /loast endpoint. When the `task` query parameter is equal to `/loast`, a request body is added to the request so that only the `positive` suite is run. The request body is added at this point so that a request body won't need to be added to the original GET request. A unit test was added as well.


mheusgit (approved):
Nothing looks out of place

PR #283, opened 2023-05-03
API-23430/Slack dashboard URLs

ediiotero:
Story: [API-23430](https://vajira.max.gov/browse/API-23430)

Update Slack URLs to match changes on the portal to include the env in the URL.


src/config.ts
```diff
@@ -104,7 +111,7 @@ const config = {
   LOAST_VERSION: 'unknown',
   LOAST_URL_PATTERN:
     process.env.LOAST_URL_PATTERN ||
-    'https://hub.lighthouse.va.gov/provider-dashboard/apis/{API_NAME}/versions/{VERSION}/loast',
+    `https://hub.lighthouse.va.gov/provider-dashboard/apis/{API_NAME}/${portalEnv}/versions/{VERSION}/loast`,
```
mheusgit, line 114: The dashboard URLs were originally designed to closely match the actual request on DocServer side.  I prefer if the env path param proceeds the rest.

https://hub.lighthouse.va.gov/provider-dashboard/${portalEnv}/apis/{API_NAME}/versions/{VERSION}/loast

src/config.ts
```diff
@@ -53,6 +58,13 @@ const ENV_METADATAS: { [env: string]: EnvironmentMetadata[] } = {
 };
 
 const env = process.env.OAS_ENV || 'unknown';
+
+let portalEnv = env;
+
+if (env === EnvType.DEV || env === 'unknown') {
+  portalEnv = PortalEnv.QA;
```
mheusgit, line 65: The logic does not appear to support links to DocServer staging or prod

ediiotero: I might be missing something but `portalEnv` is set to the current `env` that could be staging, dev, or prod. The `If` is only there to change it from dev to qa to be in line with the links on the portal.

mheusgit: I misread it.  Your good then

mheusgit (dismissed):
Looks good

dsteeve (approved):
I like the "dashboardEnvContext" change.  I think this aligns better with the future for it to be DEV.  

I realize the actual docserver context is QA (which is our DEV) -- because blue.{env}.lighthouse.va.gov doesn't have a DEV.  However the future will be LHDI which does have a DEV and a QA but that QA is more like STAGING.  LHDI won't have a STAGING, so thus I think DEV is a better choice here.  

LOL! I don't think any sane person can read that above paragraph.

dsteeve:
> I like the "dashboardEnvContext" change. I think this aligns better with the future for it to be DEV.
> 
> I realize the actual docserver context is QA (which is our DEV) -- because blue.{env}.lighthouse.va.gov doesn't have a DEV. However the future will be LHDI which does have a DEV and a QA but that QA is more like STAGING. LHDI won't have a STAGING, so thus I think DEV is a better choice here.
> 
> LOL! I don't think any sane person can read that above paragraph.

Actually, think this value needs to remain as 'qa' for now -- and change that to 'dev' as a separate ticket.

dsteeve (changes requested):
--QA is an env highly baked into the provider dashboard at  the moment, so will make that change to DEV as a separate effort.

dsteeve (approved):
looks good!

PR #284, opened 2023-05-08
[API-26426] Allow healthcheckUrl to be set back to null

jeremyhwu:
Fixing a bug where `healthcheckUrl` could not be set back to null. Now, when `healthcheckUrl` is excluded from an update, it is set to null. This change has also been applied to the `label` field.


mheusgit:
Delay performing merge until after this upcoming MR is completed.

mheusgit:
MR was completed.  This PR can proceed.

mheusgit (approved):
Looks good.  I see healthcheck was changed to match elsewhere in DocServer

PR #286, opened 2023-05-10
API-26012/  doc server  implement gateway service caching

mheusgit:
- Updated Axios version from 0.21.1 to 1.3.4
- Installed axios-cache-interceptor & node-cron.  New modules trigger bi-daily caching of gateway entities
- GatewayServiceConfig, APIVersionService, and NotificationService request handling moved to a new HTTPClient to centralize axios logic
- Implemented controller & endpoint '/cache/gateway'.  Logic depends on a new CachedHTTPClient to handle caching
- ServiceInjector no longer handles gateway, Slack, or AWS configurations instead its dealt with in the relevant service


christineyost (commented):
I read a couple places that the `node-cron schedule` function doesn't work with async functions. We might need to consider instead using `node-schedule` or the existing `lighthouse-platform-tasks` solution.

I also wonder if we should update the Docserver OAS as part of a separate ticket again because I think we have several changes that are missing at this point

docs/docserver_oas.json
```diff
@@ -920,17 +984,21 @@
                   "type": "string",
                   "example": "https://cirith.ungol/v0/web.openapi.json"
                 },
+                "healthcheckUrl": {
+                  "type": "null",
```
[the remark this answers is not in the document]
mheusgit: Its autogenerated

docs/docserver_oas.json
```diff
@@ -644,6 +693,9 @@
                   "oasUrl": {
                     "type": "string"
                   },
+                  "healthcheckUrl": {
+                    "type": "null"
+                  },
```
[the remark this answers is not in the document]
mheusgit: Its autogenerated.  The tool realizes there is property to pass but didn't detect a test passing a value

src/app.ts
```diff
@@ -218,3 +223,20 @@ function setupUncaughtErrorFallbacks(): void {
...
+  logger.info(`Prepopulating Gateway Cache on startup`);
+  await injector.gatewayService.refreshCache();
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+  cron(
+    cfg.SCHEDULER_GATEWAY_REFRESH,
+    async () => {
+      logger.info(`Scheduled Job / Refresh Cache triggered`);
+      await injector.gatewayService.refreshCache();
+    },
```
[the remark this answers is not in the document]
mheusgit: Yes.  The logic received a number of test.  I'll check for the exact problem it has with async.  End of day the cron doesn't need to return a value.  I bet that is where some DEVs ran into trouble.

docs/docserver_oas.json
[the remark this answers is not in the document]
mheusgit: For future reference only the summarys, descriptions, security declarations, and a 2 or 3 param declarations in docserver_oas_base.json are not autogenerated.  Everything else is from the tool.

src/app.ts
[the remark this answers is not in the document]
mheusgit: After further digging around I see async used in a number of places for this module.  There is only one comment in Stackoverflow saying otherwise.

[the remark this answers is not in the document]
mheusgit: Looked over this further with @christineyost .  

The type were added but the module devs did not write this package.  There are subtle differences.  From looking at the code using a async should work but trying to use a value from the returned function probably cannot be done.  Current usage does not expect this anyhow.

mheusgit:
PR should be in better shape

src/app.ts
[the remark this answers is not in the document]
christineyost: I think this is better. 

To fix the remaining eslint disable we could do a PR on the `definitelytyped` github repo adding the Promise<any> type as a return type of the function. I saw an [equivalent PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60104) on the `node-schedule` types. Definitely not needed for this change, but it might be nice to sometime look though the eslint disables we have and see if we can fix any

christineyost (commented):
Looks good to me! 

Not sure if @dsteeve might want to take a look?

mheusgit:
package-lock.json probably needs to be regenerated.  Merging it is no good

mheusgit:
As expected it created build errors.

src/gateway/gateway.service.ts
```diff
@@ -172,10 +159,35 @@ class GatewayService {
...
+
   public async healthCheck(): Promise<ServiceHealthCheckResponse> {
     try {
       const url = '/services/';
-      const response = await this.client.get<AxiosResponse>(url);
+      const response = await this.client.get<AxiosResponse>(
+        url,
+        CachedHTTPClient.directRequest(),
+      );
       if (response.status !== 200) {
```
dsteeve, line 191: a better error here might give the status it did receive to assist in troubleshooting

src/utils/cachedHTTPClient.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  // Config overrides/replaces long lived cached request
+  //  Large TTL acts as a fallback in case scheduler fails to refresh cache daily
+  protected static dailyCache(): CacheRequestConfig {
+    return {
+      cache: {
+        override: true,
+        ttl: 1000 * 60 * 60 * 24 * 7, // Week in milliseconds
```
dsteeve, line 70: this is interesting and I recommend some documentation in our quokka team page for docserver.  this could cause confusion down the road when things begin to move slow after a week because cacher wasn't working and then all the cache is deleted.

test/notification/notification.service.test.ts
```diff
@@ -95,6 +97,24 @@ const oas = {
 } as unknown as OASModel;
 const postPath = '/api/chat.postMessage';
 
+it('NotivicationService.defaultConfig() returns error on missing values', () => {
```
dsteeve, line 100: typo NotificationService.

src/gateway/gateway.service.ts
```diff
@@ -1,67 +1,54 @@
...
-  logger.info(error.message);
-  return Promise.reject(error);
-};
-
-class GatewayService {
-  public client: AxiosInstance;
-  constructor(kongConfig: KongConfig) {
-    this.client = axios.create({
-      baseURL: `${kongConfig.host}`,
+class GatewayService extends CachedHTTPClient {
```
dsteeve, line 26: at first glance, it feels weird to extend an HTTP client here.  Seems like that is more a "has a" relationship vs "is a".  Wondering why this choice?

src/notification/notification.service.ts
```diff
@@ -23,62 +14,66 @@ import logger from '../utils/logger';
...
+  baseURL: string;
+  token: string;
+  options: WebAPISlackOptions;
+}
 
-  if (body.blocks && body?.blocks[0]?.text) {
-    title = body.blocks[0].text.text;
-  }
+export default class SlackService
+  extends HTTPClient
```
dsteeve, line 26: again, seems weird to extend an HTTPClient for a Service.

src/api/version/apiVersion.service.ts
```diff
@@ -1,63 +1,19 @@
...
-  return Promise.reject(error);
-};
-
-class APIVersionService {
-  public client: AxiosInstance;
+class APIVersionService extends HTTPClient {
   public entityManager: EntityManager;
   private portalStoreService: PortalStoreService;
   constructor(orm: MikroORM, portalStoreService: PortalStoreService) {
-    this.client = axios.create({
```
dsteeve, line 44: definitely like moving this code out and just bringing in the httpclient.

dsteeve (changes requested):
Overall, I like the revamp and cleanup done here, but had a couple questions.

src/cache/cache.controller.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  recacheGateway: RequestHandler = async (request, response, next) => {
+    try {
+      await this.gatewayService.refreshCache();
+      response.sendStatus(200);
+    } catch (err) {
+      next(err);
+    }
+  };
+}
```
dsteeve, line 27: one of my comments got deleted somehow. 
curious if refreshCache will totally delete all cache and rebuild?  OR do we need a way to delete the old cache? 
past experience with caching is "refresh" did more of a replace of existing cache -- but if one item in the cache wasn't replaced (because it didn't exist in the data coming back anymore) - that one item sticks around until its TTL comes due.  

will that happen with this?  I need time to test that.

https://axios-cache-interceptor.js.org/guide/invalidating-cache

src/gateway/gateway.service.ts
mheusgit: This comment is tied to a section of code that was untouched

src/utils/cachedHTTPClient.ts
mheusgit: I could swear this was already documented.  Going to check

src/gateway/gateway.service.ts
mheusgit: Several services were already acting as HTTPClients.  This PR simply made the relationship more explicit.

src/notification/notification.service.ts
mheusgit: See above

src/api/version/apiVersion.service.ts
mheusgit: Code following the previous convention is doomed in the long term.

src/cache/cache.controller.ts
mheusgit: No, its on a URL by URL basis in CachedHTTPClient.refreshPaginatedCache()

Both the "override" and "staleIfError" flags are used to instruct axios-cache-interceptor to give desired behavior.  
From the website:
stalteIfError: Enables cache to be returned if the response comes with an error, either by invalid status code, network errors and etc.

override: This option bypasses the current cache and always make a new http request. This will not delete the current cache, it will just replace the cache when the response arrives.

Both settings were tested to confirm they work as desired.

mheusgit: ttl is used at the fallback of last resort.  In this case it has the extra long time to live

dsteeve: yep, just wondering - and it sounds like it isn't feasible - if we should have a "trashAllCache" in the event that stale data from cache is returning and we want to start over.  Sounds like a restart of the server would do the trick?

src/gateway/gateway.service.ts
dsteeve: I would say that services were using HTTPClients and were not An HTTPClient. I would like it to have a client and not be extending from a client.  Reason being, I can see a close future where there is a baseService class that owns (not extends) the HttpClient -- and we can just extend from the baseService class.

src/gateway/gateway.service.ts
dsteeve: ok, I will create a separate ticket.

src/utils/cachedHTTPClient.ts
dsteeve: are you saying it was documented in the ticket?  I was asking for documentation that was part of the ReadMe that caching of http cache exists and will expire based on the ttl configured in cachedHttpClient.ts

src/gateway/gateway.service.ts
dsteeve: I created a separate ticket for this, but since this introduced the httpClient and the cachedHTTPClient, I feel like it should be corrected as part of this ticket.

src/cache/cache.controller.ts
mheusgit: To clear or "trash" the cache you call CachedHTTPClient.refreshPaginatedCache() method or the new "POST:/cache/gateway" endpoint.

src/gateway/gateway.service.ts
mheusgit: Before you go down that route you might want to consider how you'll support overriding the client methods & interceptors.  Previously logic dealt with this by redundancies.

mheusgit: Sounds like this is less of a class structure issue and more the verbiage of "client" causing confusion.

dsteeve: no, that is not my concern.

dsteeve: will go with current state with promise of future conversations over concerns.

src/api/version/apiVersion.service.ts
dsteeve: will go with current state with promise of future conversations over concerns -- 

One recommendation is to move this httpClient out of the utils package and create an "http" package.  Then the interceptors can live there -- and when creating a client you could use default interceptors or a special crafted one like NotificationService uses. Then these interceptors could get reused. 

I see a future where clients might be graphQL -- tying a service business code to http could be limiting.

mheusgit:
PR merge must wait until after upcoming MR is completed

dsteeve (approved):

mheusgit:
MR has been performed and no longer blocks this PR

PR #288, opened 2023-05-15
API-26329/ DocServer: Add custom user agent string to request

ediiotero:
Story: [API-26329](https://vajira.max.gov/browse/API-26329)

Adding header 'User-Agent' for request Docserver makes to outside systems. 
Using `ProductName`/`ProductVersion`/`Environment` for User-Agent String.


test/loast/loast.manager.integration.ts
```diff
@@ -89,7 +90,7 @@ beforeEach(async () => {
   await loastSuiteService.createSuite(loastSuiteModelNoCredentialsJson);
   apiVersionModel = await apiVersionManager.createAPIVersion(
     apiModel,
-    apiVersionRequestWithLoast as APIVersion,
+    apiVersionRequestWithLoast as unknown as APIVersion,
```
christineyost, line 93: We shouldn't need `as unknown as APIVersion` here. Instead the `apiVersionRequestWithLoast` fixture should be updated so that the `gatewayConfig` property is changed to `gatewayConfigs` - it should be an array instead of an object, just like in the `apiVersionRequest` fixture just above it

ediiotero: Totally missed that. Thanks.

christineyost (commented):
Looks good to me! Sorry I didn't think of it before, but I have one more thought. It might be nice to store the user agent string in the cfg itself, then if we want to change the format for any reason we can do it in one place instead of needing to change it for all the different clients. Then each client could add `'User-Agent': cfg.USER_AGENT` or something like that

christineyost (dismissed):
I think we want to hold off merging since we don't have AWS access, but these changes look good to me!

src/config.ts
```diff
@@ -1,6 +1,9 @@
 //config.ts is intended to provide a centralized place to retrieve properties and common environment settings.
 //This should also contain the default overrides if a config is missing rather than each location in code having to perform defaulting...
 import { EnvironmentMetadata } from './gateway/gateway.interface';
+import fs from 'fs';
+
+const productVersion = fs.readFileSync('VERSION').toString().split('-')[0];
```
dsteeve, line 6: what happens here if perhaps some day the VERSION file is missing?

ediiotero: Good question. Should I add a condition if the version file is missing?

dsteeve (approved):
Looks good!! I like the new test too -- that will help for sure.

PR #289, opened 2023-05-17
API-26073/  docserver  jest nesting cleanup loast test

mheusgit:
- Flattened and cleaned Jest test under LOAST test area
- Moved reusable assets under fixtures were possible
- Switched manual mocking to spyOn() to allow easy cleanup instead of using beforeEach()


dsteeve:
I brought this branch down and ran it.  I think these changes are good and a great improvement! Wondering if we should fix the running of tests that broke a couple weeks back.  npm run test:unit and npm run test:integration report some issues yet. I put my findings in the Jira ticket.

dsteeve (approved):
Brought down branch, rebuilt & re-ran all tests.  Looking great!

PR #294, opened 2023-06-07
API-26329 - Copy VERSION file to prod Docker image

christineyost:
Update Dockerfile so VERSION file gets copied to the prod Docker image


dsteeve (approved):
looks correct

PR #298, opened 2023-06-20
[API-275152] Increase max length for displayName to 100

jeremyhwu:
Increasing the max length for API displayName field from 40 to 100.


mheusgit (approved):
Looks good

PR #299, opened 2023-06-21
API-27494/  migrate docserver to use aws sdk v3

mheusgit:
- Updated AWS references to follow V3 SDK convention of calling clients "@aws-sdk/*".  Import of "aws-sdk" is discontinued
- paramstore.service.ts logic required some migration.  New client changed method signature which prevented promise when callback provided
- Importing @aws-sdk/types to prevent TS errors within AWS node_modules
- Updated AWS modules to 3.350.0


dsteeve:
looks good Marc. 

 In reviewing this, I happened to notice that there is a reference to aws-sdk/util-stream-node as well, which is technically v3, so the warning message went away, but technically, it sounds like we should move away from that as well. 
https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_util_stream_node.html

want a new ticket for that?

mheusgit:
Updated to use the newer home for util-stream module

dsteeve (approved):
looks good!

PR #301, opened 2023-06-27
[API-27047] Add 'differs' property when S3 metadata is out of sync

jeremyhwu:
Published property now returns an ENUM which now includes a `differs` property. `differs` is returned when the metadata published in S3 is out of sync with newly generated metadata.


src/api/metadata/metadata.service.ts
```diff
@@ -38,30 +69,7 @@ class MetadataService {
 
   public async generateMetadata(api: APIModel): Promise<Metadata> {
```
dsteeve, line 70: 2 functions with different names work, but it would be possible to have "generateMetadata" with an optional apiVersions parameter -- and if they didn't supply it, you could go get it -- might be a better alternative?  

 I like the idea of breaking up the logic to decide if you have versions and go get them - separate from actual generating the metadata.  Seems easier to test too.  
 
 Basically one public function of generateMetadata with optional versions param supplied.  With one private function to format results.

dsteeve (changes requested):

src/api/metadata/metadata.service.ts
jeremyhwu: Having an optional `apiVersions` parameter on the function is what I wanted to do originally as well, seemed cleaner.

However, doing it that way creates a circular dependency. The metadata service calls `this.apiVersionManager.getAPIVersionsForAPI(api)` to retrieve the API Version list, and requires an instance of the `apiVersionManager` in the constructor. Since I'm trying to use the `generateMetadata` function in the `apiVersionManager` as well, I had to move the code out into a public function that isn't a class method of the metadata service. Otherwise, I wouldn't be able to import the function into the manager.

jeremyhwu: I did a bit of a refactor using Christine's suggestions and `generateMetadata` now has an optional `apiVersions` parameter. I also created a private function to format the metadata results as suggested above.

christineyost (commented):
`npm run build` resulted in some build errors. Those should be fixed along with these changes.

The APIVersionManager refactor looks good to me though! Also tested locally and the `published` property is set on responses as expected.

christineyost (approved):
Looks good to me!

dsteeve (approved):
ran build locally and reviewed latest changes.  I like the refactor, I think we are on a better track now.

PR #305, opened 2023-07-06
API-26017/ DocServer: Add Sandbox Kong Support - 2

mheusgit:
- Updated environment configs to support sandbox gateway service


mheusgit:
PR was used for a pipeline test and is no longer needed.

PR #315, opened 2023-07-25
[API-28107] Add validation for URL inputs using new Hostname enum

jeremyhwu:


src/api/version/apiVersion.schema.ts
```diff
@@ -6,6 +6,13 @@ import APIVersionStatusEnum, {
   APIVersionSecurityEnum,
   APIVersionSecurityOAuthEnum,
 } from './apiVersion.enum';
+import HostnameEnum from '../../hostname/hostname.enum';
+
+const urlPattern = new RegExp(
```
mheusgit, line 11: This pattern will be needed in multiple locations.  Best to move it to a utility

src/api/version/apiVersion.schema.ts
```diff
@@ -35,8 +42,8 @@ export const apiVersionMutateSchema = Joi.object()
       oasPath: Joi.string().required(),
       suites: Joi.array().required().items(Joi.string()).min(1).unique(),
     }),
-    oasUrl: Joi.string().required(),
-    healthcheckUrl: Joi.string(),
+    oasUrl: Joi.string().required().pattern(urlPattern, { name: 'URL' }),
```
mheusgit, line 45: What error message does this generate?  Pretty sure you need to override the message with something more helpful.

src/hostname/hostname.enum.ts
```diff
@@ -0,0 +1,9 @@
+// Enum containing all valid hostnames for URL inputs
+enum HostnameEnum {
```
mheusgit, line 2: Move to "input validation" utility

src/api/version/apiVersion.schema.ts
mheusgit: I see a test with message "oasUrl" with value "https://theshire.gov/api/hobbits/v2/docs/healthcheck" fails to match the URL pattern"

src/api/version/apiVersion.schema.ts
```diff
@@ -35,8 +42,8 @@ export const apiVersionMutateSchema = Joi.object()
       oasPath: Joi.string().required(),
```
mheusgit, line 42: URL pattern and hostname check needed here

mheusgit:
Ticket ask suggest using "enum list" which is what this PR does.  

Most of the unit test changes could have been avoided if a Node environment property was used with a CSV list of permitted hostnames.  You could then add a new ALLOWED_HOSTNAMES to "/test/globalSetup.ts" with the host the unit test were using.

That approach could avoid future code changes as new hostnames are added but that is out of scope for now.

src/api/version/apiVersion.schema.ts
jeremyhwu: I split up the URL validation into two different patterns, should improve this:
* the first one checks the hostname
* the second one is a generic URL pattern

Now the error is clearer and states whether it is caused by an invalid hostname or invalid URL format.

src/api/version/apiVersion.schema.ts
jeremyhwu: I split up the URL validation into two different patterns, should improve this:
* the first one checks the hostname
* the second one is a generic URL pattern

Now the error is clearer and states whether it is caused by an invalid hostname or invalid URL format.

src/hostname/hostname.enum.ts
jeremyhwu: Moved into urlValidation util file

src/api/version/apiVersion.schema.ts
jeremyhwu: I added a new validation pattern for server URLs that allows the `{version}` format to pass

src/api/version/apiVersion.schema.ts
jeremyhwu: Moved into `urlValidation` util file

PR #317, opened 2023-07-28
API-28296 - use generated GITHUB_TOKEN in workflow

christineyost:
Change GitHub Action for security alert notifications to use the auto-generated GITHUB_TOKEN


dsteeve (approved):
appears to be correct as it follows github docs to do it this way.

PR #318, opened 2023-07-28
API-26072 fix broken test.  healthcheck import coming in undefined

dsteeve:
fix broken test because import wasn't being brought in


christineyost (approved):

christineyost (changes requested):
Changing oneRingServer's api fixture import would fix this. But a better fix might be moving the healthcheck constant to a fixture file. It seems like more of an APIVersion fixture than something related to the oneRingServer

dsteeve:
> Changing oneRingServer's api fixture import would fix this. But a better fix might be moving the healthcheck constant to a fixture file. It seems like more of an APIVersion fixture than something related to the oneRingServer

yes, that is a better long term fix.

mheusgit (approved):

christineyost (approved):

stevelong00 (commented):
Looks good!

PR #320, opened 2023-07-31
API-28610: remove dockerfile copy in favor of base image dealing with…

dsteeve:
remove dockerfile copy in favor of base image dealing with the certs as necessary. 
This change should help with the upgrade of certs on the load balancers.


mramirez-va (approved):

mheusgit (approved):
I see both parts (SSL setup and environment setup) ended up being removed.

PR #321, opened 2023-08-01
[API-28475] Add `publishIt` column to `api_version` table

jeremyhwu:
Add `publishIt` column to `api_version` table with a default value of `true`


src/api/version/apiVersion.schema.ts
```diff
@@ -44,6 +44,7 @@ export const apiVersionMutateSchema = Joi.object()
       .required()
       .valid(...Object.values(APIVersionStatusEnum)),
     refresh: Joi.boolean().required(),
+    publishIt: Joi.boolean().required(),
```
christineyost, line 47: This might be the way we want to do this, but just wanted to point out this will make this field required when creating and updating an APIVersion, which would break Provider Dashboard until the new checkbox is added. Is that how we should implement this? @dsteeve @mheusgit

christineyost (commented):
Saw there are some compile errors when I ran `npm run build` in the test files. Those should be fixed before merging. I'll create a ticket to look into adding that step to our Jenkins build

PR #322, opened 2023-08-01
API-28296 - test removing dependabot check from GitHub action

christineyost:
Test removing dependabot check from GitHub action


dsteeve:
approving to see if it is just the dependabot having issues with the auth token.

dsteeve (approved):
lets try it

